PR #1, opened 2017-04-28
New dynamic logging using go-kit

melekes:
https://github.com/tendermint/tendermint/pull/473

### DIFF
--- a/common/service.go
+++ b/common/service.go
@@ -3,7 +3,7 @@ package common
 import (
 	"sync/atomic"
 
-	"github.com/tendermint/log15"
+	"github.com/tendermint/tmlibs/log"
 )
 
 type Service interface {
@@ -19,6 +19,8 @@ type Service interface {
 	IsRunning() bool
 
 	String() string
+
+	SetLogger(log.Logger)
 }
 
 /*
@@ -64,7 +66,7 @@ Typical usage:
 	}
 */
 type BaseService struct {
-	log     log15.Logger
+	Logger  log.Logger
 	name    string
 	started uint32 // atomic
 	stopped uint32 // atomic
@@ -74,27 +76,31 @@ type BaseService struct {
 	impl Service
 }
 
-func NewBaseService(log log15.Logger, name string, impl Service) *BaseService {
-	return &BaseService{
-		log:  log,
-		name: name,
-		Quit: make(chan struct{}),
-		impl: impl,
+func NewBaseService(logger log.Logger, name string, impl Service) *BaseService {
+	if logger == nil {
+		logger = log.NewNopLogger()
 	}
+
+	return &BaseService{
+		Logger: logger,
+		name:   name,
+		Quit:   make(chan struct{}),
+		impl:   impl,
+	}
+}
+
+func (bs *BaseService) SetLogger(l log.Logger) {
+	bs.Logger = l
 }
 
 // Implements Servce
 func (bs *BaseService) Start() (bool, error) {
 	if atomic.CompareAndSwapUint32(&bs.started, 0, 1) {
 		if atomic.LoadUint32(&bs.stopped) == 1 {
-			if bs.log != nil {
-				bs.log.Warn(Fmt("Not starting %v -- already stopped", bs.name), "impl", bs.impl)
-			}
+			bs.Logger.Error(Fmt("Not starting %v -- already stopped", bs.name), "impl", bs.impl)
 			return false, nil
 		} else {
-			if bs.log != nil {
-				bs.log.Info(Fmt("Starting %v", bs.name), "impl", bs.impl)
-			}
+			bs.Logger.Info(Fmt("Starting %v", bs.name), "impl", bs.impl)
 		}
 		err := bs.impl.OnStart()
 		if err != nil {
@@ -104,9 +110,7 @@ func (bs *BaseService) Start() (bool, error) {
 		}
 		return true, err
 	} else {
-		if bs.log != nil {
-			bs.log.Debug(Fmt("Not starting %v -- already started", bs.name), "impl", bs.impl)
-		}
+		bs.Logger.Debug(Fmt("Not starting %v -- already started", bs.name), "impl", bs.impl)
 		return false, nil
 	}
 }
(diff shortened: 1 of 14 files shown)


log/logger.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"fmt"
+
+	kitlog "github.com/go-kit/kit/log"
+)
+
+// Logger is what any Tendermint library should take.
+type Logger interface {
+	Debug(msg string, keyvals ...interface{}) error
```
ebuchman: why do these return errors? are we expected to handle them?

log/tm_logger.go
```diff
@@ -0,0 +1,77 @@
...
+	srcLogger kitlog.Logger
+}
+
+// NewTmTermLogger returns a logger that encodes msg and keyvals to the Writer
+// using go-kit's log as an underlying logger and our custom formatter. Note
+// that underlying logger could be swapped with something else.
+func NewTmLogger(w io.Writer) Logger {
+	// Color by level value
+	colorFn := func(keyvals ...interface{}) term.FgBgColor {
+		for i := 0; i < len(keyvals)-1; i += 2 {
```
ebuchman: so, is this happening on every call to the logger? it scans all the keys looking for the log level ?

log/tm_logger.go
```diff
@@ -0,0 +1,77 @@
...
+	case "debug":
+		return &tmLogger{level.NewFilter(l.srcLogger, level.AllowDebug())}
+	case "error":
+		return &tmLogger{level.NewFilter(l.srcLogger, level.AllowError())}
+	default:
+		panic(fmt.Sprintf("Unexpected level %v, expect either \"info\" or \"debug\" or \"error\"", lvl))
+	}
+}
+
+// Info logs a message at level Info on the logrus logger.
```
ebuchman: logrus?

log/tmfmt_logger.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+			continue
+		}
+
+		if keyvals[i] == msgKey {
+			msg = keyvals[i+1].(string)
+			continue
+		}
+
+		restKeyvals = append(restKeyvals, keyvals[i], keyvals[i+1])
```
ebuchman: this oughta slow us down!

log/logger.go
melekes: No, we are not. Still it's good to have in case you want to check why logger does not produce output or write tests. 

Originally from https://github.com/go-kit/kit/blob/master/log/log.go#L11
Error could be like https://github.com/go-kit/kit/blob/master/log/json_logger.go#L34 json encoding error

log/tm_logger.go
melekes: yeah, that's probably not good for performance. I might know the way how to fix that (have index keyval always first)

log/tmfmt_logger.go
melekes: I will create a benchmark. Still it should be way faster than log15

melekes:
```
BenchmarkTmLoggerSimple-2                 500000              2270 ns/op             816 B/op       17 allocs/op
BenchmarkTmLoggerContextual-2             500000              2900 ns/op            1328 B/op       25 allocs/op
BenchmarkTmfmtLoggerSimple-2             1000000              1873 ns/op             368 B/op         11 allocs/op
BenchmarkTmfmtLoggerContextual-2          500000              2401 ns/op             704 B/op         16 allocs/op
```

need to write the same benchmark for log15 to be able to compare

log/tmfmt_logger.go
melekes: https://github.com/tendermint/tmlibs/pull/1#issuecomment-298072400

melekes: wait. i know how to fix this too

log/tm_logger.go
melekes: no, no difference

```
BenchmarkTmLoggerSimple-2                 500000              2720 ns/op            1424 B/op         31 allocs/op
BenchmarkTmLoggerContextual-2            1000000              2193 ns/op            1040 B/op         20 allocs/op
BenchmarkTmfmtLoggerSimple-2             1000000              1659 ns/op             272 B/op          9 allocs/op
BenchmarkTmfmtLoggerContextual-2         1000000              2017 ns/op             480 B/op         13 allocs/op
PASS
ok      github.com/tendermint/tmlibs/log        7.361s
root@f15c047b4b19:/go/src/github.com/tendermint/tmlibs/log# go test -bench .
BenchmarkTmLoggerSimple-2                 500000              2726 ns/op            1424 B/op         31 allocs/op
BenchmarkTmLoggerContextual-2            1000000              2143 ns/op            1040 B/op         20 allocs/op
BenchmarkTmfmtLoggerSimple-2             1000000              1701 ns/op             272 B/op          9 allocs/op
BenchmarkTmfmtLoggerContextual-2         1000000              2082 ns/op             480 B/op         13 allocs/op
```

melekes: but will do this optimization anyway

melekes:
Latest benchmark:

```
BenchmarkTmLoggerSimple-2                1000000              1764 ns/op             720 B/op       15 allocs/op
BenchmarkTmLoggerContextual-2             500000              2196 ns/op            1040 B/op       20 allocs/op
BenchmarkTmfmtLoggerSimple-2             1000000              1733 ns/op             272 B/op        9 allocs/op
BenchmarkTmfmtLoggerContextual-2          500000              2067 ns/op             480 B/op       13 allocs/op
```

common/service.go
```diff
@@ -64,7 +66,7 @@ Typical usage:
 	}
 */
 type BaseService struct {
-	log     log15.Logger
+	Logger  log.Logger
```
ebuchman: Does it need to be exposed?

melekes: Yes. We could have a method `GetLogger`, but field is faster and looks cleaner. See https://github.com/tendermint/abci/pull/78

log/tmfmt_logger.go
```diff
@@ -0,0 +1,116 @@
...
+		enc.Encoder = logfmt.NewEncoder(&enc.buf)
+		return &enc
+	},
+}
+
+type tmfmtLogger struct {
+	w io.Writer
+}
+
+// NewTmFmtLogger returns a logger that encodes keyvals to the Writer in
```
ebuchman: Doesn't match the function, NewTmfmtLogger. should be Fmt

ebuchman: actually the standard is to make uppercase abbreviations. so probably should be NewTMFmtLogger

melekes: I thought it depends on a filename. ie. `tm_fmt` => `TmFmt`, `tmfmt` => `Tmfmt`

> actually the standard is to make uppercase abbreviations

ok

PR #3, opened 2017-05-01
cli

ebuchman:
I think this needs some more discussion before merging, but I do like having some common CLI setup in one place that all tools can use.

Note this is coming from Frey's work in go-crypto:
https://github.com/tendermint/go-crypto/blob/develop/cmd/README.md
https://github.com/tendermint/go-crypto/blob/develop/cmd/common.go

1) should every tool really have output/encoding flags? Maybe this needs to be optional and configured with another arg to `PrepareMainCmd`
2) Current ENV var is TMROOT. This would change it to TM_ROOT (I think). Is that what we want ?
3) Other ?

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -1,0 +1,111 @@
+package cli
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	data "github.com/tendermint/go-wire/data"
+	"github.com/tendermint/go-wire/data/base58"
+)
+
+const (
+	RootFlag     = "root"
+	OutputFlag   = "output"
+	EncodingFlag = "encoding"
+)
+
+func PrepareMainCmd(cmd *cobra.Command, envPrefix, defautRoot string) func() {
+	cobra.OnInitialize(func() { initEnv(envPrefix) })
+	cmd.PersistentFlags().StringP(RootFlag, "r", defautRoot, "root directory for config and data")
+	cmd.PersistentFlags().StringP(EncodingFlag, "e", "hex", "Binary encoding (hex|b64|btc)")
+	cmd.PersistentFlags().StringP(OutputFlag, "o", "text", "Output format (text|json)")
+	cmd.PersistentPreRunE = multiE(bindFlags, setEncoding, validateOutput, cmd.PersistentPreRunE)
+	return func() { execute(cmd) }
+}
+
+// initEnv sets to use ENV variables if set.
+func initEnv(prefix string) {
+	// env variables with TM prefix (eg. TM_ROOT)
+	viper.SetEnvPrefix(prefix)
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	viper.AutomaticEnv()
+}
+
+// Execute calls cmd.Execute and exits if there is an error.
+func Execute(cmd *cobra.Command) {
+	if err := cmd.Execute(); err != nil {
+		fmt.Println(err)
+		os.Exit(-1)
+	}
+}
+
+type wrapE func(cmd *cobra.Command, args []string) error
+
+func multiE(fs ...wrapE) wrapE {
+	return func(cmd *cobra.Command, args []string) error {
+		for _, f := range fs {
+			if f != nil {
+				if err := f(cmd, args); err != nil {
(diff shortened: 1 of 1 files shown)


adrianbrink:
I agree that I common CLI setup would be great, since it will make our CLI's behave almost the same for all projects which is great from a user perspective.

What is the reason behind changing from TMROOT to TM_ROOT?

ethanfrey:
TM_ROOT comes from viper, how they bind environmental variables with prefixes.

`viper.SetEnvPrefix("TM")` means `TM_ROOT` is bound to the `--root` flag, `TM_NODE` is bound to the `--node` flag etc.  This is viper standard.  I don't know how hard it would be to change, and if so, would be applied to all env variables.

I also agree `PrepareMainCmd` should be split into two commands.  One that handles root, config file, (maybe debug), etc... And a second call that does this, and adds the output and encoding flags.  Maybe call that `PrepareClientCmd` and let `PrepareMainCmd` be the simpler version.  I guess controlling printing errors with `--debug` or `TM_DEBUG/TMDEBUG` could go there as well?

ethanfrey:
Background: https://godoc.org/github.com/spf13/viper#SetEnvPrefix

The line where this is called (not so easy to override...):
https://github.com/spf13/viper/blob/master/viper.go#L307

rigelrozanski:
@ebuchman I've moved some of the common cobra aux functionality that I built for basecoin (which can now be used by basecoin-plugins and light-client). maybe I'll PR it to this PR, add all the cobra utilities in one place - right now its in the branch `common-cli`

cli/cli.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/tendermint/go-wire/data/base58"
+)
+
+const (
+	RootFlag     = "root"
+	OutputFlag   = "output"
+	EncodingFlag = "encoding"
+)
+
+func PrepareMainCmd(cmd *cobra.Command, envPrefix, defautRoot string) func() {
```
rigelrozanski: Can we have a couple sentences at the top of this command describing a usage scenario? Also add option for ExecuteWithDebug? (see PR from `basecoin-cli`)

cli/cli.go
```diff
@@ -0,0 +1,111 @@
...
+func Execute(cmd *cobra.Command) {
+	if err := cmd.Execute(); err != nil {
+		fmt.Println(err)
+		os.Exit(-1)
+	}
+}
+
+type wrapE func(cmd *cobra.Command, args []string) error
+
+func multiE(fs ...wrapE) wrapE {
```
rigelrozanski: We should add a small comment describing intended usage here too I believe

ethanfrey:
Oops, yesterday trying to get out quick for my gf's bday, I wrote this directly on develop. Commits b92bd8f and d4ab967 include a refactored version that takes into account Bucky's concerns. I will add some comments as requested by Rigel now.

Sorry to skip this PR.

melekes:
Also, do not forget to update glide when you add new deps, so others won't be spending time figuring out why glide isn't working (see https://github.com/tendermint/tmlibs/pull/5/files)

ethanfrey:
Good point.  Will push with this soon.

ethanfrey:
Okay, lots of fixes pushed in 305cc24

The glide is properly updated.  There are actual test cases for the cli functionality. (Yes, everything can be tested).  And I fixed some bugs in the first implementation of --home.

cli/cli.go
```diff
@@ -0,0 +1,111 @@
...
+	EncodingFlag = "encoding"
+)
+
+func PrepareMainCmd(cmd *cobra.Command, envPrefix, defautRoot string) func() {
+	cobra.OnInitialize(func() { initEnv(envPrefix) })
+	cmd.PersistentFlags().StringP(RootFlag, "r", defautRoot, "root directory for config and data")
+	cmd.PersistentFlags().StringP(EncodingFlag, "e", "hex", "Binary encoding (hex|b64|btc)")
+	cmd.PersistentFlags().StringP(OutputFlag, "o", "text", "Output format (text|json)")
+	cmd.PersistentPreRunE = multiE(bindFlags, setEncoding, validateOutput, cmd.PersistentPreRunE)
+	return func() { execute(cmd) }
```
rigelrozanski: should be `Execute(cmd)` ?

ethanfrey: This is totally different in develop.... I'm going to close this pr.

PR #4, opened 2017-05-02
add basecoin cli commands

rigelrozanski:


ethanfrey:
The `StringVar` commands and such, that bind to one variable don't seem to work to well with viper.  Best is just to make a flag, and access with `viper.GetString`, so it let's you check config, env vars, and flags and default vals in one swoop.

The addition to add debug for Execute, I like.  I think @ebuchman had some issue with this and log-levels.

And yes, the `cmd.SilentUsage = true` line is very good.

I will add the SilentUsage stuff in the current code, and wait for feedback from Bucky about how to handle debug vs log-level. I mean, there is also a difference for server commands (which are long-running and write a log-file as a secondary info), and "interactive" commands, where the main output is on standard output (the actual response) and doesn't need to be mixed with logging info so much.

Let's have a longer discussion on this soon.

rigelrozanski:
Yeah, I was following that we're maybe replacing --debug with log levels, can make sense - should we have a breakout meeting discussing this soon? 

Okay, I can do a swap out of this functionality in favour of non-binding flags.

ethanfrey:
Hmmm.. I went ahead and added --debug support in 501f706 for now.

Let's see how to rework with log-level later.  But the tests should still work.

ethanfrey:
Just check out develop.  I took what I could, since there were a number of other changes since. Let me know what you think.  Also, look at the tests.  I want to test all the cli functions now...

rigelrozanski:
@ethanfrey Okay looks really good to incorporate debug as you've done here, Do you think we just skip the quick flag imports all-together? I'm on the fence, doesn't really save that much room/maybe just an over-abstraction? if you agree then let's just close this PR

rigelrozanski:
More on this, the only advantage I was thinking earlier for being able to register duplicate sets of flags between commands, but I just realized we can register entire flagsets with `cmd.Flags.AddFlagSet(`

PR #5, opened 2017-05-03
Testing logger

melekes:
Helper function to be used in testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
-*.swp
-*.swo
+vendor
+.glide
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONEY: all test install get_vendor_deps ensure_tools
+.PHONY: all test install get_vendor_deps ensure_tools
 
 GOTOOLS = \
 	github.com/Masterminds/glide
--- a/cli/setup.go
+++ b/cli/setup.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+
 	data "github.com/tendermint/go-wire/data"
 	"github.com/tendermint/go-wire/data/base58"
 )
@@ -24,7 +25,7 @@ func PrepareBaseCmd(cmd *cobra.Command, envPrefix, defautRoot string) func() {
 	cobra.OnInitialize(func() { initEnv(envPrefix) })
 	cmd.PersistentFlags().StringP(RootFlag, "r", defautRoot, "DEPRECATED. Use --home")
 	cmd.PersistentFlags().StringP(HomeFlag, "h", defautRoot, "root directory for config and data")
-	cmd.PersistentPreRunE = multiE(bindFlags, cmd.PersistentPreRunE)
+	cmd.PersistentPreRunE = concatCobraCmdFuncs(bindFlagsLoadViper, cmd.PersistentPreRunE)
 	return func() { execute(cmd) }
 }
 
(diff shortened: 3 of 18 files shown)


ethanfrey:
Here are some questions I have on the design... not necessarily problems, but things I noticed.
First on the `develop` branch...

The `Logger` interface is nice and lean, but there are two quasi-hidden functions there that might desired to be upgraded to first class citizens:

`log.With()` is super useful, but limits the implementations to what it supports. I'd suggest adding `With(keyvals ...interface{})` to the Logger interface, to make Logger more self-contained.

`*tmLogger.WithLevel()` is another useful function, but requires that we have a `*tmLogger`.  I don't know if this breaks the transparency of the Logger interface.  Or if it is only supposed to be called in the setup phase, when we work with the real struct, before passing it into the system that will write logs.  In that case, maybe just mention that it is intended to be used in setup.

ethanfrey:
Okay, now `feature/testing-logger` branch.

I see you moved `WithLevel` into a wrapper function that has to switch on the type, which makes it harder to add my own logger interface in my own package and expect it to work (will likely panic when With is called, without anything I can do).

Here is a realistic usecase.  I want to programatically test the log output in certain cases... I would then create a `MockLogger` class that appends log messages unformatted to a slice, and public accessors to read it in the case.  I could call a function with `MockLogger` and make sure some notes were printed out.  Or at least assert no-errors, or yes, this kind of error happened inside.

You don't always want this, but I have found MockLoggers and MockWriters useful in many high-level tests.

However, I would have to modify `With` and possibly `WithLevel` to support it.  This seems more brittle than adding it to the interface. (Yes, I could pass a bytes.Buffer as the io.Writer, but then I get the formatted data).

Also, I notice the comment on `NewTMFmtLogger` that the writer must be safe for concurrent go-routines if it will be used by concurrent go-routines.  With that in mind, it might make sense to add a helper function like:

```
func SafeifyWriter(w io.Writer) io.Writer {
  return SafeWriter{w}
}
```

and `SafeWriter` uses a mutex to make any writer concurrency safe.

ethanfrey:
The last point is formatting.

The log level seems to affect the coloring in the output, which is great.  There are two points that might make it nicer.

One, that it would be cool to have another formatter that outputs json for programatic parsing, like logrus does.  One json object per line and that makes it easier to parse and send it off to elk or other such...

Second, it would be nice to have some flag to print the whole stack trace for errors.  That means, normally with "%v" or `err.Error()`, and with some flag (`--debug` or such) you use "%+v".  I am not sure how it would work, but definitely nice to use for debugging, to print full stack traces when debugging, without changing the code.  This makes full use of `pkg/errors`

melekes:
> log.With() is super useful, but limits the implementations to what it supports. I'd suggest adding With(keyvals ...interface{}) to the Logger interface, to make Logger more self-contained.

I think you are right. I will move this method to the interface.

> *tmLogger.WithLevel() is another useful function, but requires that we have a *tmLogger. I don't know if this breaks the transparency of the Logger interface. Or if it is only supposed to be called in the setup phase, when we work with the real struct, before passing it into the system that will write logs. In that case, maybe just mention that it is intended to be used in setup.

Will add the comment.

> I see you moved WithLevel into a wrapper function that has to switch on the type, which makes it harder to add my own logger interface in my own package and expect it to work (will likely panic when With is called, without anything I can do).

It will be used only during setup (`run_node` command) with our logger. So I guess it is ok.

melekes:
> https://github.com/tendermint/tmlibs/pull/5#issuecomment-299199481

this could be done in a separate PR

ebuchman:
Closing. Sorry for any trouble but lets keep it all in https://github.com/tendermint/tmlibs/pull/7

PR #7, opened 2017-05-05
Log

ebuchman:
Combines https://github.com/tendermint/tmlibs/pull/1 and https://github.com/tendermint/tmlibs/pull/5. See the discussion in https://github.com/tendermint/tmlibs/pull/5

### DIFF
--- a/common/service.go
+++ b/common/service.go
@@ -3,7 +3,7 @@ package common
 import (
 	"sync/atomic"
 
-	"github.com/tendermint/log15"
+	"github.com/tendermint/tmlibs/log"
 )
 
 type Service interface {
@@ -19,6 +19,8 @@ type Service interface {
 	IsRunning() bool
 
 	String() string
+
+	SetLogger(log.Logger)
 }
 
 /*
@@ -64,7 +66,7 @@ Typical usage:
 	}
 */
 type BaseService struct {
-	log     log15.Logger
+	Logger  log.Logger
 	name    string
 	started uint32 // atomic
 	stopped uint32 // atomic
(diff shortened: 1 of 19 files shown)


common/service.go
```diff
@@ -19,6 +19,8 @@ type Service interface {
 	IsRunning() bool
 
 	String() string
+
+	SetLogger(log.Logger)
```
ebuchman: Can u remind me why the constructors are taking loggers if we have this method ?

log/filter.go
```diff
@@ -0,0 +1,97 @@
...
+	for _, option := range options {
+		option(l)
+	}
+	return l
+}
+
+type filter struct {
+	next          Logger
+	allowed       level
+	errNotAllowed error
```
ebuchman: what makes this not nil?

common/service.go
melekes: they should not. but in single cases where we log something in a constructor, there is no other way

log/filter.go
melekes: option below ErrNotAllowed

log/filter.go
```diff
@@ -0,0 +1,97 @@
...
+func AllowAll() Option {
+	return AllowDebug()
+}
+
+// AllowDebug allows error, warn, info and debug level log events to pass.
+func AllowDebug() Option {
+	return allowed(levelError | levelInfo | levelDebug)
+}
+
+// AllowInfo allows error, warn and info level log events to pass.
```
ebuchman: warn? (and above)

log/filter.go
```diff
@@ -0,0 +1,97 @@
...
+
+func allowed(allowed level) Option {
+	return func(l *filter) { l.allowed = allowed }
+}
+
+// ErrNotAllowed sets the error to return from Log when it squelches a log
+// event disallowed by the configured Allow[Level] option. By default,
+// ErrNotAllowed is nil; in this case the log event is squelched with no
+// error.
+func ErrNotAllowed(err error) Option {
```
ebuchman: I see. so callers need to call this if they want the err to not be nil

melekes: `log.NewFilter(logger, log.ErrNotAllowed(errors.New("hey")))`

https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

ebuchman: love it

PR #10, opened 2017-05-06
Close file before renaming it

melekes:
this might fix our windows bug https://github.com/tendermint/tendermint/issues/444

https://github.com/Netflix-Skunkworks/go-jira/commit/0980f8e1972a942e505b1939935adf0f7a71f387

Backport of https://github.com/tendermint/go-autofile/pull/3

### DIFF
--- a/autofile/group.go
+++ b/autofile/group.go
@@ -224,15 +224,17 @@ func (g *Group) RotateFile() {
 	g.mtx.Lock()
 	defer g.mtx.Unlock()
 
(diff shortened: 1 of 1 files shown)


ebuchman:
How can we replicate the bug and test the fix? Have you tested this with tendermint WAL rollover on linux ?

melekes:
> How can we replicate the bug and test the fix? 

do we have a windows machine?

> Have you tested this with tendermint WAL rollover on linux/mac ?

yes (linux)

PR #11, opened 2017-05-08
Squash module keyvals if multiple keyvals were provided

melekes:
**last keyvalue wins**

**Pros**:
- still fast
```
BenchmarkTMLoggerSimple-2                 300000              5953 ns/op             721 B/op       15 allocs/op
BenchmarkTMLoggerContextual-2             200000              6463 ns/op            1073 B/op       21 allocs/op
BenchmarkTMFmtLoggerSimple-2              200000              5427 ns/op             273 B/op        9 allocs/op
BenchmarkTMFmtLoggerContextual-2          200000              6455 ns/op             481 B/op       13 allocs/op

BenchmarkTMLoggerSimple-2                 200000              6083 ns/op             752 B/op         17 allocs/op
BenchmarkTMLoggerContextual-2             200000              6221 ns/op            1104 B/op         23 allocs/op
BenchmarkTMFmtLoggerSimple-2              300000              4892 ns/op             273 B/op          9 allocs/op
BenchmarkTMFmtLoggerContextual-2          200000              5874 ns/op             481 B/op         13 allocs/op
```

**Cons**:
- `tmfmt_logger` now knows about the `module` key and that there could be multiple values.

The same thing could be implemented via a new wrapper (similar to filter API `NewFilter(logger, ...)`) which has a `map[interface{}]interface{}` to hold unique keyvalue pairs https://gist.github.com/melekes/a662c8f7809626f1911a20c5672a6228. But having a map will result in twice more memory!

### DIFF
--- a/log/tm_logger.go
+++ b/log/tm_logger.go
@@ -10,7 +10,8 @@ import (
 )
 
 const (
-	msgKey = "_msg" // "_" prefixed to avoid collisions
+	msgKey    = "_msg" // "_" prefixed to avoid collisions
+	moduleKey = "module"
 )
 
 type tmLogger struct {
--- a/log/tmfmt_logger.go
+++ b/log/tmfmt_logger.go
@@ -51,13 +51,15 @@ func (l tmfmtLogger) Log(keyvals ...interface{}) error {
 
 	lvl := "none"
 	msg := "unknown"
-	lvlIndex := -1
-	msgIndex := -1
+	module := "unknown"
+
+	// indexes of keys to skip while encoding later
+	excludeIndexes := make([]int, 0)
 
 	for i := 0; i < len(keyvals)-1; i += 2 {
 		// Extract level
 		if keyvals[i] == kitlevel.Key() {
-			lvlIndex = i
+			excludeIndexes = append(excludeIndexes, i)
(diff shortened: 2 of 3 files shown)


ebuchman:
> tmfmt_logger now knows about the module key and that there could be multiple values.

What makes this a con? Are you saying its breaking an abstraction layer? Shouldn't we be preparing for the logger to potentially have more access to structured info anyways? For instance if we eventually thread the `context` objects through everything. 

Also, did we decide against concatenating the modules with `/`?  Looks like you're just using the last one here.

melekes:
> Also, did we decide against concatenating the modules with /?

everyone said they don't care. We can do the nested version no problem.

> What makes this a con? Are you saying its breaking an abstraction layer?

Not sure. Before there were `_msg` and `error` keys, but they are sort of built in (well, not exactly as they are coming from tm_logger, but still). Now we start making assumptions about user-space keyvalues.

> Shouldn't we be preparing for the logger to potentially have more access to structured info anyways? For instance if we eventually thread the context objects through everything.

Could you elaborate on this? Are you thinking of smth like `context.Logger.GetWithFields()`?

melekes:
On the second thought, it is our format and we can apply any rules we want.

PR #17, opened 2017-05-16
Allow custom levels for the keyvals in filter

melekes:


ethanfrey:
It would be great if you could pull in `tendermint/cmd/tendermint/flags` from https://github.com/tendermint/tendermint/pull/501 into the cli directory to help parsing.

PR #19, opened 2017-05-22
copy log level parsing from tendemint

melekes:
API change due to me not wanting `flags` package to depend on
tendermint's config package.

Refs https://github.com/tendermint/tendermint/issues/504

### DIFF
--- a/cli/flags/log_level.go
+++ b/cli/flags/log_level.go
@@ -1,0 +1,86 @@
+package flags
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"github.com/tendermint/tmlibs/log"
+)
+
+const (
+	defaultLogLevelKey = "*"
+)
+
+// ParseLogLevel parses complex log level - comma-separated
+// list of module:level pairs with an optional *:level pair (* means
+// all other modules).
+//
+// Example:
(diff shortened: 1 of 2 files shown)


ethanfrey:
Looks good to me.  I would enhance the tests though. It doesn't show how default and other packages work, as everything writes to mempool in all cases.

ethanfrey:
Btw, why did you put them in a subdir of cli? There isn't much in cli.  Is it the import of go-wire/data that bothers you?

Huh, on second thought, maybe that is not good....

ethanfrey:
Okay, take a look at the test changes I made. They make it clearer how the different modules play together.  But maybe there is something untested now?

melekes:
> Btw, why did you put them in a subdir of cli? There isn't much in cli. Is it the import of go-wire/data that bothers you?

`flags.ParseLogLevel` just looks cleaner. Should we move these files into cli?

melekes:
> Okay, take a look at the test changes I made. They make it clearer how the different modules play together. But maybe there is something untested now?

Like them! I think we are lacking a test case when there is no "module" key. I will add it.

ethanfrey:
> flags.ParseLogLevel just looks cleaner. Should we move these files into cli?

Or maybe I should move the other flags parsing from cli into cli/flags. It does look nice. I would just get @ebuchman  opinion as flags is a standard name often taken by other packages. In any case, we should be consistent.

Issue #16, opened 2017-05-15
Flowrate tests fail non-deterministically


ebuchman:
https://circleci.com/gh/tendermint/tmlibs/59

melekes:
Sometimes `time.Sleep` takes more than 100ms

```
clock before sleeping: 0.78
Sleeping for 100ms
clock after sleeping: 0.9
```

melekes:
Don't know how to fix this except not comparing durations. But rates are also different when durations are.

```
                expected: {true 2017-05-23 13:22:55.4 +0000 UTC 400ms 0s 80 4 200 200 200 200 20 100ms 80.000%}
                got     : {true 2017-05-23 13:22:55.4 +0000 UTC 420ms 0s 80 4 200 197 190 200 20 100ms 80.000%}
```

melekes:
https://github.com/tendermint/tmlibs/pull/20#issuecomment-303534296

melekes:
Fixed in https://github.com/tendermint/tmlibs/pull/20/commits/5f20b3323e6afa49eb7c681a953611af08f43206. If it is a bad way and you can think of something else, please tell me

adrianbrink:
Is there a specific reasoning behind these two constants:
```
const maxDeviationForDuration = 50 * time.Millisecond
const maxDeviationForRate int64 = 50
```

If so, we should probably document it.

melekes:
@adrianbrink no. it's basically `{max seen deviation * 2}`

PR #20, opened 2017-05-23
[flowrate] refactor clock functions (Refs #16)

melekes:
this commit does not fix the original bug

### DIFF
--- a/flowrate/io_test.go
+++ b/flowrate/io_test.go
@@ -91,7 +91,7 @@ func TestReader(t *testing.T) {
 	}
 	for i, s := range status {
 		if !reflect.DeepEqual(&s, &want[i]) {
-			t.Errorf("r.Status(%v) expected %v; got %v", i, want[i], s)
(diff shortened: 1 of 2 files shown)


ethanfrey:
This is still a WIP?  Or did you fix the bug?

I had this issue too.... rather annoying the randomly failing tests... Somewhere there is an unexpected clock tick... guess sleep is not precise to the millisecond.

melekes:
@ethanfrey https://github.com/tendermint/tmlibs/issues/16#issuecomment-303396143

ethanfrey:
Fair enough.  I took a look at the code myself and left scratching my head.  But random fails is no good... maybe we just disable those tests?? Use some "approx equal" method?

melekes:
I wonder when did it break? maybe something had changed in Go

> Use some "approx equal" method?

I guess we can compare that the time is within range and rates are greater than X

adrianbrink:
@melekes Is there anything blocking this PR from being merged?

PR #21, opened 2017-05-23
[log] tracing logger

melekes:


log/tracing_logger.go
```diff
@@ -0,0 +1,55 @@
...
+func (l *tracingLogger) With(keyvals ...interface{}) Logger {
+	return &tracingLogger{next: l.next.With(formatErrors(keyvals)...)}
+}
+
+func formatErrors(keyvals []interface{}) []interface{} {
+	newKeyvals := make([]interface{}, len(keyvals))
+	copy(newKeyvals, keyvals)
+	for i := 0; i < len(newKeyvals)-1; i += 2 {
+		if err, ok := newKeyvals[i+1].(stackTracer); ok {
+			newKeyvals[i+1] = fmt.Sprintf("%+v", err)
```
ethanfrey: It is actually nice to leave the original error in some cases... but the printout should be as you test.

I would wrap it with another error that prints stuff out. Will make another commit you can accept or reject.

Why?  Some handlers might want access to the error itself and the StackTrace, not just in text form.

ethanfrey:
I like the work.  I would like to enhance it a bit more.

ethanfrey:
I added two commits to show my thoughts.  Feel free to revert them if you don't agree. Especially b9acb2a

log/tracing_logger.go
```diff
@@ -0,0 +1,76 @@
...
+func (l *tracingLogger) With(keyvals ...interface{}) Logger {
+	return &tracingLogger{next: l.next.With(formatErrors(keyvals)...)}
+}
+
+func formatErrors(keyvals []interface{}) []interface{} {
+	newKeyvals := make([]interface{}, len(keyvals))
+	copy(newKeyvals, keyvals)
+	for i := 0; i < len(newKeyvals)-1; i += 2 {
+		if err, ok := newKeyvals[i+1].(stackTracer); ok {
+			newKeyvals[i+1] = tracedError{err}
```
ethanfrey: Yeah, this is probably clearer this way...

melekes:
> I added two commits to show my thoughts. Feel free to revert them if you don't agree. Especially b9acb2a

I now remember us talking about this :) thanks!

squashed commits. hope you don't mind

adrianbrink:
@ebuchman Still requires your review? :-)

PR #22, opened 2017-06-05
IsHex and StripHex

rigelrozanski:


common/string.go
```diff
@@ -22,3 +23,22 @@ func LeftPadString(s string, totalLength int) string {
...
+		_, err := hex.DecodeString(s[2:])
+		if err != nil {
+			return false
+		}
+		return true
+	}
+	return false
+}
+
+func StripHex(s string) string {
```
melekes: missing doc

common/string.go
```diff
@@ -22,3 +23,22 @@ func LeftPadString(s string, totalLength int) string {
 	}
 	return s
 }
+
+// Returns true for non-empty hex-string prefixed with "0x"
```
melekes: golint would have told you doc should be in a form of "IsHex ..."

rigelrozanski: Thanks just installed golint addressed in PR#23

common/string.go
rigelrozanski: Thanks just installed golint addressed in PR#23

PR #24, opened 2017-06-06
common date parsing

rigelrozanski:
parsing dates and date ranges

### DIFF
--- a/common/date.go
+++ b/common/date.go
@@ -1,0 +1,67 @@
+package common
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// ParseDate parses a date string of the format YYYY-MM-DD
+func ParseDate(date string) (t time.Time, err error) {
+
+	//get the time of invoice
+	str := strings.Split(date, "-")
+	var ymd = []int{}
+	for _, i := range str {
+		j, err := strconv.Atoi(i)
+		if err != nil {
+			return t, err
+		}
+		ymd = append(ymd, j)
+	}
+	if len(ymd) != 3 {
+		return t, fmt.Errorf("Bad date parsing, not 3 segments") //never stack trace
+	}
+	if ymd[1] < 1 || ymd[1] > 12 {
+		return t, fmt.Errorf("Month not between 1 and 12") //never stack trace
+	}
+	if ymd[2] > 31 {
+		return t, fmt.Errorf("Day over 31") //never stack trace
+	}
+
+	t = time.Date(ymd[0], time.Month(ymd[1]), ymd[2], 0, 0, 0, 0, time.UTC)
+
+	return t, nil
+}
+
+// ParseDateRange parses a date range string of the format start:end
+//   where the start and end date are of the format YYYY-MM-DD.
+//   The parsed dates are *time.Time and will return nil pointers for
+//   unbounded dates, ex:
+//   unbounded start:	:2000-12-31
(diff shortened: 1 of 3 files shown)


common/date.go
```diff
@@ -0,0 +1,67 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// ParseDate parses a date string of the format YYYY-MM-DD
+func ParseDate(date string) (t time.Time, err error) {
```
melekes: aren't there any existing functions? https://play.golang.org/p/0Q1ChGzGwJ

rigelrozanski: I had a bunch of trouble getting this to work... really confusing with the month day being represented by the 02 and 01 respectively. BUT THANKS. this makes sense... Do we still want to include this library for date range parsing? Also nice to just have the layout predefined here I think... I'll update the code

rigelrozanski: like why on earth is the layout have to be January 2, 2006?? haha

rigelrozanski: okay update here

common/date.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// ParseDate parses a date string of the format YYYY-MM-DD
+func ParseDate(date string) (t time.Time, err error) {
+	layout := "2006-01-02" //this represents YYYY-MM-DD
```
melekes: why not a constant? then we could remove this func altogether

rigelrozanski: This function is used outside of the ParseDateRange function too, so it should probably exist. We can make layout a constant but I don't think it makes a difference, your call given that this function needs to exist either way

melekes: > This function is used outside of the ParseDateRange function too

where exactly?

rigelrozanski: I'm using in trackomatron

ebuchman: If we make the layout an exposed constant, others can just call `time.Parse(cmn.TimeLayour, date)`, and then we don't need this function. I don't mind it too much though

common/date_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}{
+		{"2015-12-31", date, true},
+		{"2015-31-12", date, false},
+		{"12-31-2015", date, false},
+		{"31-12-2015", date, false},
+	}
+
+	for _, test := range testDates {
+		parsed, err := ParseDate(test.dateStr)
+		switch test.errNil {
```
ebuchman: No need for a switch/case if there's only two cases. If statement is better

common/date.go
```diff
@@ -0,0 +1,43 @@
...
+	return time.Parse(layout, date)
+}
+
+// ParseDateRange parses a date range string of the format start:end
+//   where the start and end date are of the format YYYY-MM-DD.
+//   The parsed dates are *time.Time and will return nil pointers for
```
ebuchman: Why do we return pointers? Is it so we can use nil? Is the zero time and IsZero not sufficient?

ebuchman:
Changelog please

common/date.go
rigelrozanski: Okay, don't really mind, implemented ebuchman's suggestion

common/date_test.go
rigelrozanski: Thanks, changed

common/date.go
rigelrozanski: Oh great, yes, I agree that IsZero makes more sense

PR #25, opened 2017-06-06
CLI Execute Error prints to stderr

rigelrozanski:
instead of stdout as before, very useful for bash processing/error checking for CLI testing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+*.swp
 vendor
 .glide
--- a/cli/helper.go
+++ b/cli/helper.go
@@ -68,21 +68,18 @@ func RunCaptureWithArgs(cmd Executable, args []string, env map[string]string) (s
 	}()
 
 	// copy the output in a separate goroutine so printing can't block indefinitely
-	outC := make(chan string)
-	go func() {
-		var buf bytes.Buffer
-		// io.Copy will end when we call wOut.Close() below
-		io.Copy(&buf, rOut)
-		outC <- buf.String()
-	}()
-
-	errC := make(chan string)
-	go func() {
-		var buf bytes.Buffer
-		// io.Copy will end when we call wErr.Close() below
-		io.Copy(&buf, rErr)
-		errC <- buf.String()
-	}()
+	copyStd := func(reader *os.File) *(chan string) {
+		stdC := make(chan string)
+		go func() {
+			var buf bytes.Buffer
+			// io.Copy will end when we call reader.Close() below
+			io.Copy(&buf, *reader)
+			stdC <- buf.String()
+		}()
+		return stdC
+	}
+	outC := copyStd(&rOut)
+	errC := copyStd(&rErr)
 
 	// now run the command
 	err = RunWithArgs(cmd, args, env)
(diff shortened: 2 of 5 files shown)


cli/helper.go
```diff
@@ -57,28 +57,39 @@ func RunWithArgs(cmd Executable, args []string, env map[string]string) error {
 // RunCaptureWithArgs executes the given command with the specified command line args
```
melekes: we should update the docs too

cli/helper.go
```diff
@@ -57,28 +57,39 @@ func RunWithArgs(cmd Executable, args []string, env map[string]string) error {
...
 	// now run the command
 	err = RunWithArgs(cmd, args, env)
 
 	// and grab the stdout to return
-	w.Close()
-	output = <-outC
-	return output, err
+	wOut.Close()
+	wErr.Close()
+	stdout = <-outC
```
melekes: what if command only writes to stderr? will this code block forever?

rigelrozanski: I tried to make this happen with one pipe if that's what you're getting at not sure it's possible

rigelrozanski: Every example I've seen used recieves "" if standard output is empty so it doesn't lock

cli/helper.go
rigelrozanski: done

cli/helper.go
melekes: ok, good

.gitignore
```diff
@@ -1,2 +1,3 @@
+*.swp
```
melekes: https://robots.thoughtbot.com/global-gitignore

rigelrozanski: Thanks for the tip, just implemented on my machine

cli/helper.go
```diff
@@ -54,31 +54,43 @@ func RunWithArgs(cmd Executable, args []string, env map[string]string) error {
...
 		var buf bytes.Buffer
-		// io.Copy will end when we call w.Close() below
-		io.Copy(&buf, r)
+		// io.Copy will end when we call wOut.Close() below
+		io.Copy(&buf, rOut)
 		outC <- buf.String()
 	}()
 
+	errC := make(chan string)
+	go func() {
```
ebuchman: Can we clean up the code dup here ?

cli/helper.go
```diff
@@ -54,31 +54,43 @@ func RunWithArgs(cmd Executable, args []string, env map[string]string) error {
...
-// stdout along with any error returned from cmd.Execute()
-func RunCaptureWithArgs(cmd Executable, args []string, env map[string]string) (output string, err error) {
-	old := os.Stdout // keep backup of the real stdout
-	r, w, _ := os.Pipe()
-	os.Stdout = w
+// RunCaptureWithArgs executes the given command with the specified command
+// line args and environmental variables set. It returns string fields
+// representing output written to stdout and stderr, additionally any error
+// from cmd.Execute() is also returned
+func RunCaptureWithArgs(cmd Executable, args []string, env map[string]string) (stdout, stderr string, err error) {
```
ebuchman: Note this is breaking

ebuchman:
Changelog please

cli/helper.go
rigelrozanski: added to changelog

cli/helper.go
rigelrozanski: Okay updated, let me know if sufficient, not sure we can compress this anymore

PR #28, opened 2017-06-22
New pubsub package plus the query subpackage

melekes:
Implemented as a part of https://github.com/tendermint/tendermint/pull/548

### DIFF
--- a/pubsub/example_test.go
+++ b/pubsub/example_test.go
@@ -1,6 +1,7 @@
 package pubsub_test
 
 import (
+	"context"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -16,9 +17,11 @@ func TestExample(t *testing.T) {
 	s.Start()
 	defer s.Stop()
 
+	ctx := context.Background()
 	ch := make(chan interface{}, 1)
-	err := s.Subscribe("example-client", query.MustParse("abci.account.name=John"), ch)
+	err := s.Subscribe(ctx, "example-client", query.MustParse("abci.account.name=John"), ch)
 	require.NoError(t, err)
-	s.PublishWithTags("Tombstone", map[string]interface{}{"abci.account.name": "John"})
+	err = s.PublishWithTags(ctx, "Tombstone", map[string]interface{}{"abci.account.name": "John"})
+	require.NoError(t, err)
 	assertReceive(t, "Tombstone", ch)
 }
--- a/pubsub/example_test.go
+++ b/pubsub/example_test.go
@@ -17,8 +17,8 @@ func TestExample(t *testing.T) {
 	defer s.Stop()
 
 	ch := make(chan interface{}, 1)
-	s.Subscribe("example-client", query.MustParse("abci.account.name=John"), ch)
-	err := s.PublishWithTags("Tombstone", map[string]interface{}{"abci.account.name": "John"})
+	err := s.Subscribe("example-client", query.MustParse("abci.account.name=John"), ch)
 	require.NoError(t, err)
+	s.PublishWithTags("Tombstone", map[string]interface{}{"abci.account.name": "John"})
 	assertReceive(t, "Tombstone", ch)
 }
--- a/pubsub/example_test.go
+++ b/pubsub/example_test.go
@@ -1,0 +1,24 @@
+package pubsub_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/tendermint/tmlibs/log"
+	"github.com/tendermint/tmlibs/pubsub"
+	"github.com/tendermint/tmlibs/pubsub/query"
+)
+
+func TestExample(t *testing.T) {
+	s := pubsub.NewServer()
+	s.SetLogger(log.TestingLogger())
+	s.Start()
+	defer s.Stop()
+
+	ch := make(chan interface{}, 1)
+	s.Subscribe("example-client", query.MustParse("abci.account.name=John"), ch)
+	err := s.PublishWithTags("Tombstone", map[string]interface{}{"abci.account.name": "John"})
+	require.NoError(t, err)
+	assertReceive(t, "Tombstone", ch)
+}
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -15,6 +15,7 @@ package pubsub
 
 import (
 	"errors"
+	"time"
 
 	cmn "github.com/tendermint/tmlibs/common"
 )
@@ -28,8 +29,10 @@ const (
 	shutdown
 )
 
+const subscribeTimeout = 10 * time.Millisecond
+
 var (
-	ErrorOverflow = errors.New("Server overflowed")
+	ErrorOverflow = errors.New("server overflowed")
 )
 
 type cmd struct {
@@ -51,7 +54,8 @@ type Query interface {
 type Server struct {
 	cmn.BaseService
 
-	cmds chan cmd
+	cmds    chan cmd
+	cmdsCap int
 }
 
 // Option sets a parameter for the server.
@@ -68,9 +72,8 @@ func NewServer(options ...Option) *Server {
 		option(s)
 	}
 
-	if s.cmds == nil { // if BufferCapacity was not set, create unbuffered channel
-		s.cmds = make(chan cmd)
-	}
+	// if BufferCapacity option was not set, the channel is unbuffered
+	s.cmds = make(chan cmd, s.cmdsCap)
 
 	return s
 }
@@ -82,40 +85,49 @@ func NewServer(options ...Option) *Server {
 func BufferCapacity(cap int) Option {
 	return func(s *Server) {
 		if cap > 0 {
-			s.cmds = make(chan cmd, cap)
+			s.cmdsCap = cap
 		}
 	}
 }
 
-// Subscribe returns a channel on which messages matching the given query can
-// be received. If the subscription already exists old channel will be closed
-// and new one returned.
-func (s *Server) Subscribe(clientID string, query Query, out chan<- interface{}) {
-	s.cmds <- cmd{op: sub, clientID: clientID, query: query, ch: out}
+// Returns capacity of the internal server's queue.
+func (s Server) BufferCapacity() int {
+	return s.cmdsCap
 }
 
-// Unsubscribe unsubscribes the given client from the query.
+// Subscribe returns a channel on which messages matching the given query can
+// be received. If the subscription already exists old channel will be closed
+// and new one returned. Error will be returned to the caller if the server is
+// overflowed.
+func (s *Server) Subscribe(clientID string, query Query, out chan<- interface{}) error {
+	select {
+	case s.cmds <- cmd{op: sub, clientID: clientID, query: query, ch: out}:
+		return nil
+	case <-time.After(subscribeTimeout):
+		return ErrorOverflow
+	}
+}
+
+// Unsubscribe unsubscribes the given client from the query. Blocking.
 func (s *Server) Unsubscribe(clientID string, query Query) {
 	s.cmds <- cmd{op: unsub, clientID: clientID, query: query}
 }
 
-// Unsubscribe unsubscribes the given channel.
+// Unsubscribe unsubscribes the given channel. Blocking.
 func (s *Server) UnsubscribeAll(clientID string) {
 	s.cmds <- cmd{op: unsub, clientID: clientID}
 }
 
-// Publish publishes the given message.
-func (s *Server) Publish(msg interface{}) error {
-	return s.PublishWithTags(msg, make(map[string]interface{}))
+// Publish publishes the given message. Blocking.
+func (s *Server) Publish(msg interface{}) {
+	s.PublishWithTags(msg, make(map[string]interface{}))
 }
 
 // PublishWithTags publishes the given message with a set of tags. This set of
 // tags will be matched with client queries. If there is a match, the message
-// will be sent to a client.
-func (s *Server) PublishWithTags(msg interface{}, tags map[string]interface{}) error {
-	pubCmd := cmd{op: pub, msg: msg, tags: tags}
-	s.cmds <- pubCmd
-	return nil
+// will be sent to a client. Blocking.
+func (s *Server) PublishWithTags(msg interface{}, tags map[string]interface{}) {
+	s.cmds <- cmd{op: pub, msg: msg, tags: tags}
 }
 
 // OnStop implements Service.OnStop by shutting down the server.
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -110,8 +110,8 @@ func (s *Server) Unsubscribe(ctx context.Context, clientID string, query Query) 
 	}
 }
 
-// Unsubscribe removes all client subscriptions. An error will be returned to
-// the caller if the context is canceled.
+// UnsubscribeAll removes all client subscriptions. An error will be returned
+// to the caller if the context is canceled.
 func (s *Server) UnsubscribeAll(ctx context.Context, clientID string) error {
 	select {
 	case s.cmds <- cmd{op: unsub, clientID: clientID}:
@@ -172,7 +172,7 @@ loop:
 				state.removeAll(cmd.clientID)
 			}
 		case shutdown:
-			for clientID, _ := range state.clients {
+			for clientID := range state.clients {
 				state.removeAll(clientID)
 			}
 			break loop
@@ -232,7 +232,7 @@ func (state *state) removeAll(clientID string) {
 		return
 	}
 
-	for q, _ := range queryMap {
+	for q := range queryMap {
 		ch := state.queries[q][clientID]
 		close(ch)
 
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -10,26 +10,13 @@
 // match, this message will be pushed to all clients, subscribed to that query.
 // See query subpackage for our implementation.
 //
-// Overflow strategies (incoming publish requests):
-//
-// 1) drop - drops publish requests when there are too many of them
-// 2) wait - blocks until the server is ready to accept more publish requests (default)
-//
 // Subscribe/Unsubscribe calls are always blocking.
-//
-// Overflow strategies (outgoing messages):
-//
-// 1) skip - do not send a message if the client is busy or slow (default)
-// 2) wait - wait until the client is ready to accept new messages
-//
 package pubsub
 
 import (
 	"errors"
-	"fmt"
 
 	cmn "github.com/tendermint/tmlibs/common"
-	"github.com/tendermint/tmlibs/log"
 )
 
 type operation int
@@ -39,13 +26,6 @@ const (
 	pub
 	unsub
 	shutdown
-)
-
-type overflowStrategy int
-
-const (
-	drop overflowStrategy = iota
-	wait
 )
 
 var (
@@ -72,20 +52,16 @@ type Server struct {
 	cmn.BaseService
 
 	cmds chan cmd
-
-	overflowStrategy   overflowStrategy
-	slowClientStrategy overflowStrategy
 }
 
 // Option sets a parameter for the server.
 type Option func(*Server)
 
 // NewServer returns a new server. See the commentary on the Option functions
-// for a detailed description of how to configure buffering and overflow
-// behavior. If no options are provided, the resulting server's queue is
-// unbuffered and it blocks when overflowed.
+// for a detailed description of how to configure buffering. If no options are
+// provided, the resulting server's queue is unbuffered.
 func NewServer(options ...Option) *Server {
-	s := &Server{overflowStrategy: wait, slowClientStrategy: drop}
+	s := &Server{}
 	s.BaseService = *cmn.NewBaseService(nil, "PubSub", s)
 
 	for _, option := range options {
@@ -108,38 +84,6 @@ func BufferCapacity(cap int) Option {
 		if cap > 0 {
 			s.cmds = make(chan cmd, cap)
 		}
-	}
-}
-
-// OverflowStrategyDrop will tell the server to drop messages when it can't
-// process more messages.
-func OverflowStrategyDrop() Option {
-	return func(s *Server) {
-		s.overflowStrategy = drop
-	}
-}
-
-// OverflowStrategyWait will tell the server to block and wait for some time
-// for server to process other messages. Default strategy.
-func OverflowStrategyWait() func(*Server) {
-	return func(s *Server) {
-		s.overflowStrategy = wait
-	}
-}
-
-// WaitSlowClients will tell the server to block and wait until subscriber
-// reads a messages even if it is fast enough to process them.
-func WaitSlowClients() func(*Server) {
-	return func(s *Server) {
-		s.slowClientStrategy = wait
-	}
-}
-
-// SkipSlowClients will tell the server to skip subscriber if it is busy
-// processing previous message(s). Default strategy.
-func SkipSlowClients() func(*Server) {
-	return func(s *Server) {
-		s.slowClientStrategy = drop
 	}
 }
 
@@ -170,17 +114,7 @@ func (s *Server) Publish(msg interface{}) error {
 // will be sent to a client.
 func (s *Server) PublishWithTags(msg interface{}, tags map[string]interface{}) error {
 	pubCmd := cmd{op: pub, msg: msg, tags: tags}
-	switch s.overflowStrategy {
-	case drop:
-		select {
-		case s.cmds <- pubCmd:
-		default:
-			s.Logger.Error("Server overflowed, dropping message...", "msg", msg, "tags", fmt.Sprintf("%v", tags))
-			return ErrorOverflow
-		}
-	case wait:
-		s.cmds <- pubCmd
-	}
+	s.cmds <- pubCmd
 	return nil
 }
 
@@ -224,7 +158,7 @@ loop:
 		case sub:
 			state.add(cmd.clientID, cmd.query, cmd.ch)
 		case pub:
-			state.send(cmd.msg, cmd.tags, s.slowClientStrategy, s.Logger)
+			state.send(cmd.msg, cmd.tags)
 		}
 	}
 }
@@ -289,20 +223,11 @@ func (state *state) removeAll(clientID string) {
 	delete(state.clients, clientID)
 }
 
-func (state *state) send(msg interface{}, tags map[string]interface{}, slowClientStrategy overflowStrategy, logger log.Logger) {
+func (state *state) send(msg interface{}, tags map[string]interface{}) {
 	for q, clientToChannelMap := range state.queries {
 		if q.Matches(tags) {
-			for clientID, ch := range clientToChannelMap {
-				switch slowClientStrategy {
-				case drop:
-					select {
-					case ch <- msg:
-					default:
-						logger.Error("Wanted to send a message, but the client is busy", "msg", msg, "tags", fmt.Sprintf("%v", tags), "clientID", clientID)
-					}
-				case wait:
-					ch <- msg
-				}
+			for _, ch := range clientToChannelMap {
+				ch <- msg
 			}
 		}
 	}
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -290,11 +290,8 @@ func (state *state) removeAll(clientID string) {
 
 func (state *state) send(msg interface{}, tags map[string]interface{}, slowClientStrategy overflowStrategy, logger log.Logger) {
 	for q, clientToChannelMap := range state.queries {
-		// NOTE we can use LRU cache to speed up common cases like query = "
-		// tm.events.type=NewBlock" and tags = {"tm.events.type": "NewBlock"}
 		if q.Matches(tags) {
 			for clientID, ch := range clientToChannelMap {
-				logger.Info("Sending message to client", "msg", msg, "client", clientID)
 				switch slowClientStrategy {
 				case drop:
 					select {
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -216,7 +216,9 @@ loop:
 				state.removeAll(cmd.clientID)
 			}
 		case shutdown:
-			state.reset()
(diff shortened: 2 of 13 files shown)


events.v2/events.go
```diff
@@ -0,0 +1,133 @@
+package events
```
melekes: rename to `pubsub`?

events.v2/events.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+}
+
+func (state *state) add(query string, ch chan interface{}) {
+	state.queries[ch] = query
+}
+
+func (state *state) send(msg interface{}, tags map[string]interface{}) {
+	for ch, query := range state.queries {
+		if match(query, tags) {
```
melekes: add a comment about cache (we can use cache to speed up common cases like `query = tm.events.type=NewBlock and tags = {"tm.events.type": "NewBlock"}`)

events.v2/events_test.go
```diff
@@ -0,0 +1,93 @@
...
+	s.Publish("Lester")
+	assert.Equal(t, "Lester", <-ch)
+
+	s.Unsubscribe(ch)
+
+	_, ok := <-ch
+	assert.False(t, ok)
+}
+
+// func Benchmark10Clients(b *testing.B)   { benchmarkNClients(10, b) }
```
melekes: uncomment!

events.v2/events_test.go
```diff
@@ -0,0 +1,93 @@
...
+func TestUnsubscribe(t *testing.T) {
+	t.Parallel()
+
+	s := events.NewServer(1)
+	s.Start()
+	defer s.Stop()
+
+	ch := s.Subscribe("")
+	s.Publish("Lester")
+	assert.Equal(t, "Lester", <-ch)
```
melekes: write a wrapper using `select` with timeout = 10ms so we don't block if a value is not there

events.v2/query/query.go
```diff
@@ -0,0 +1,215 @@
...
+	opContains
+)
+
+type operatorAndValue struct {
+	Op        operator
+	ValueType reflect.Type
+	Value     reflect.Value
+}
+
+type queryMap struct {
```
melekes: remove the map, just go over the AST from `query.peg.go` (run a benchmark to see if is not slower)

events.v2/events.go
melekes: +

pubsub/Makefile
```diff
@@ -0,0 +1,5 @@
+gen_query_parser:
```
ethanfrey: How about using a tools dependency here, like in tendermint?

```
GOTOOLS = github.com/pointlander/peg

ensure_tools:
	go get $(GOTOOLS)
```

pubsub/pubsub_test.go
```diff
@@ -0,0 +1,94 @@
...
+func TestQueries(t *testing.T) {
+	t.Parallel()
+
+	s := pubsub.NewServer(1)
+	s.Start()
+	defer s.Stop()
+
+	ch1 := s.Subscribe(q("tm.events.type=NewBlock"))
+	s.PublishWithTags("Iceman", map[string]interface{}{"tm.events.type": "NewBlock"})
+	assertReceive(t, "Iceman", ch1)
```
ethanfrey: Can you give another name in each test, so it is clear they are different.

pubsub/pubsub_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+func TestPublish(t *testing.T) {
+	t.Parallel()
+
+	s := pubsub.NewServer(1)
+	s.Start()
+	defer s.Stop()
+
+	ch := s.Subscribe(query.Empty{})
+	s.Publish("Ka-Zar")
```
ethanfrey: Again, please two different names, so we can ensure that they are received in the same order they are sent.

Also, can you test what happens when eg. 50 events are sent before the client listens?  What is the capacity of the queue and what happens when it is full?

pubsub/query/parser_test.go
```diff
@@ -0,0 +1,54 @@
...
+		{"tx.date > TIME 2013-05-03T14:45:00+07:00", true},
+		{"tx.date < TIME 2013-05-03T14:45:00-02:00", true},
+		{"tx.date <= TIME 2013-05-03T14:45:00Z", true},
+		{"tx.date >= TIME 2013-05-03T14:45:00Z", true},
+	}
+
+	for _, c := range cases {
+		_, err := query.New(c.query)
+		if c.valid {
+			assert.NoError(t, err, fmt.Sprintf("Query was '%s'", c.query))
```
ethanfrey: You can also write this:

assert.Nil(t, err, "Query '%s': %+v", c.query, err)

pubsub/query/parser_test.go
```diff
@@ -0,0 +1,54 @@
...
+		} else {
+			assert.Error(t, err, fmt.Sprintf("Query was '%s'", c.query))
```
ethanfrey: assert.NotNil(t, err, "Query '%s'", c.query)

These are the forms usually used in other test cases.  Sprintf is implied with the assert, if there are extra arguments

pubsub/query/query.go
```diff
@@ -0,0 +1,202 @@
...
+			return v.Equal(operand.Interface().(time.Time)) || v.After(operand.Interface().(time.Time))
+		case opLess:
+			return v.Before(operand.Interface().(time.Time))
+		case opGreater:
+			return v.After(operand.Interface().(time.Time))
+		case opEqual:
+			return v.Equal(operand.Interface().(time.Time))
+		}
+	case reflect.Float64:
+		v, ok := value.(float64)
```
ethanfrey: So if they query with eg. a float64, and we publish with an int, they will never match.

Fair enough, but wonder how this is clarified by the parser and serialization.  I know json `map[string]interface{}` always unserializes numbers as float64, regardless of how they were originally stored.

ethanfrey: I don't know how the parser works and all, but it would be good to have an end-to-end test to make sure matches don't fail on such an issue (maybe this is more a tendermint problem than this library)

pubsub/query/query.go
```diff
@@ -0,0 +1,202 @@
...
+		case opEqual:
+			return v == operand.Interface().(int)
+		}
+	case reflect.String:
+		v, ok := value.(string)
+		if !ok {
+			return false
+		}
+		switch op {
+		case opEqual:
```
ethanfrey: No Less or Greater on strings?  They do have a native ordering.

pubsub/query/query_test.go
```diff
@@ -0,0 +1,43 @@
...
+func TestMatches(t *testing.T) {
+	testCases := []struct {
+		s       string
+		tags    map[string]interface{}
+		err     bool
+		matches bool
+	}{
+		{"tm.events.type=NewBlock", map[string]interface{}{"tm.events.type": "NewBlock"}, false, true},
+
+		{"tx.gas > 7", map[string]interface{}{"tx.gas": 8}, false, true},
```
ethanfrey: okay, i guess this answers my questions above.  if there is a point, then it is a float.

Question: what if the json was not created in memory, but passed over the wire (from the abci app)?

pubsub/Makefile
melekes: Don't think we need it here (at least for now). the `gen_query_parser` command will be used very rarely and the whole Makefile is just one command right now.

pubsub/pubsub_test.go
melekes: You set the capacity when creating a server `pubsub.Server(1000)`. Right now the server will block if the client's queue is full. So one client can be a reason of the server freezing. I think we can provide an option to the `Server` like:

`wait` - wait indefinitely (~block)
`wait, 10 * time.Millisecond` - wait 10ms and drop
`drop` - drop

pubsub/query/query.go
melekes: > json map[string]interface{} always unserializes numbers as float64, regardless of how they were originally stored.

interesting. but ABCI app will be using protobuf to encode tags. and yes, end-to-end test is a must have

pubsub/query/query.go
melekes: Do we want that?

pubsub/pubsub_test.go
melekes: @ebuchman ^^^

ethanfrey: I support quickly dropping any messages when the queue is full.  Ideally no wait. Otherwise a few poorly implemented (or malicious) clients could halt the tendermint server.

pubsub/query/query.go
```diff
@@ -152,7 +152,13 @@ func match(tag string, op operator, operand reflect.Value, tags map[string]inter
 	case reflect.Float64:
 		v, ok := value.(float64)
 		if !ok {
-			return false
+			// if value is int, convert it to float64
```
ethanfrey: Nice, I think this may fix the json encoding issue.  Maybe lose precision?  Dunna how the casting works with floats and int.

As you mentioned above, end-to-end test is important, but I guess that belongs in tendermint itself.

ethanfrey: Also, already with two type checks, it may be easier to read a type switch

pubsub/query/query_test.go
```diff
@@ -21,6 +20,8 @@ func TestMatches(t *testing.T) {
 
 		{"tx.gas > 7", map[string]interface{}{"tx.gas": 8}, false, true},
 		{"body.weight >= 3.5", map[string]interface{}{"body.weight": 3.5}, false, true},
+		{"account.balance < 1000.0", map[string]interface{}{"account.balance": 900}, false, true},
```
ethanfrey: nice cases

pubsub/pubsub_test.go
melekes: > Otherwise a few poorly implemented (or malicious) clients could halt the tendermint server.

well, yes and no. because this is not the end client. In TM it's all up to us how to implement the end client (WSClient) https://github.com/tendermint/tendermint/blob/ec6341215e61d04892c263318a8d06313c3f6968/rpc/core/events.go#L20. There we just drop the messages if the WebSocket queue is full.

pubsub/query/query.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	switch operand.Kind() {
+	case reflect.Struct: // time
+		v, ok := value.(time.Time)
+		if !ok {
+			return false
+		}
+		switch op {
+		case opLessEqual:
+			return v.Before(operand.Interface().(time.Time)) || v.Equal(operand.Interface().(time.Time))
```
ebuchman: can we put  `operand.Interface().(time.Time)` in a variable so we only need to write it once ? same for those below

pubsub/query/query.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+	case reflect.Float64:
+		var v float64
+		switch vt := value.(type) {
+		case float64:
+			v = vt
+		case int:
+			v = float64(vt)
+		default:
+			return false
```
ebuchman: how could this happen ?

pubsub/query/query.go
```diff
@@ -0,0 +1,212 @@
...
+		case opLessEqual:
+			return v <= operand.Interface().(float64)
+		case opGreaterEqual:
+			return v >= operand.Interface().(float64)
+		case opLess:
+			return v < operand.Interface().(float64)
+		case opGreater:
+			return v > operand.Interface().(float64)
+		case opEqual:
+			return v == operand.Interface().(float64)
```
ebuchman: is equality well defined for floats ?

pubsub/query/parser_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+		{"tx.date > DATE 2013-05-03", true},
+		{"tx.date < DATE 2013-05-03", true},
+		{"tx.date <= DATE 2013-05-03", true},
+		{"tx.date >= DATE 2013-05-03", true},
+
+		{"tx.date > TIME 2013-05-03T14:45:00+07:00", true},
+		{"tx.date < TIME 2013-05-03T14:45:00-02:00", true},
+		{"tx.date <= TIME 2013-05-03T14:45:00Z", true},
+		{"tx.date >= TIME 2013-05-03T14:45:00Z", true},
```
ebuchman: can we add some `false` cases for DATE/TIME ? also would be good to add cases for ints/floats (including false cases like `43.21.1` ... or is that just a string ? ). Also false cases for things like `AN` and `TIM`, and `==` on strings ... edge-cases.

pubsub/helpers_test.go
```diff
@@ -0,0 +1,30 @@
...
+	select {
+	case actual := <-ch:
+		assert.Equal(t, expected, actual, msgAndArgs)
+	case <-time.After(10 * time.Millisecond):
+		t.Errorf("Expected to receive %v from the channel, got nothing after 10ms", expected)
+		debug.PrintStack()
+	}
+}
+
+func q(s string) *query.Query {
```
ebuchman: maybe we want to expose this functionality, like a `MustParse` ? not necessarily, just an idea.

pubsub/example_test.go
```diff
@@ -0,0 +1,21 @@
+package pubsub_test
+
+import (
+	"testing"
```
ebuchman: does this file add anything over the `pubsub_test.go` ?

pubsub/helpers_test.go
```diff
@@ -0,0 +1,30 @@
+package pubsub_test
```
ebuchman: can we just put this into `pubsub_test.go` ?

pubsub/helpers_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"github.com/tendermint/tmlibs/pubsub/query"
+)
+
+func assertReceive(t *testing.T, expected interface{}, ch chan interface{}, msgAndArgs ...interface{}) {
+	select {
+	case actual := <-ch:
+		assert.Equal(t, expected, actual, msgAndArgs)
+	case <-time.After(10 * time.Millisecond):
```
ebuchman: might be tight! maybe a time.Second is better ? Just thinking about CI servers and VMs ...

pubsub/Makefile
ebuchman: fine by me

pubsub/pubsub.go
```diff
@@ -0,0 +1,208 @@
...
+		case sub:
+			state.add(cmd.query, cmd.ch)
+		case pub:
+			state.send(cmd.msg, cmd.tags)
+		}
+	}
+}
+
+func (state *state) add(q Query, ch chan<- interface{}) {
+	state.queries[ch] = q
```
ebuchman: what if we want multiple subscriptions on a single channel ? should we add support for `OR` to enable this ? or do we make this a list of queries that is appended to ? at least I think we need to document that subsequent subscriptions on the same channel overwrite older ones (or am I wrong about this ? maybe a test for this case is warranted)

pubsub/pubsub.go
```diff
@@ -0,0 +1,208 @@
...
+		s.cmds = make(chan cmd)
+	}
+
+	return s
+}
+
+// BufferCapacity allows you to specify capacity for the internal servers
+// queue. Since the server given Y subscribers could only process X messages,
+// this option could be used to survive spikes (e.g. high amount of
+// transactions during peak hours).
```
melekes: We could run out of memory if there are too many subscribers. Otherwise, according to the Little's law https://en.wikipedia.org/wiki/Little%27s_law and queuing theory

suppose 100 subscribers
λ = 1000 events per second
W = 10ms to match event tags against all 100 subscribers

then L = 1000events/sec * 10ms = 1events/ms * 10 = 10

i.e. an average number of events in the server's queue - 10 and we need BuffersCapacity set to 10 or better 20 to accommodate spikes (like when the arrival rate is 9000 events per second, not usual 1000)

Hope that helps. Or maybe I am wrong here? Anyway, I don't think we will hit the place where we will start dropping events (especially, if we'll optimize it), but every time I design something, I think about that kind of stuff and try to design it the way so nothing super bad happens.

melekes: How can I improve my comment to reflect that?

pubsub/pubsub.go
```diff
@@ -0,0 +1,208 @@
...
+func (s *Server) Publish(msg interface{}) error {
+	return s.PublishWithTags(msg, make(map[string]interface{}))
+}
+
+// PublishWithTags publishes the given message with a set of tags. This set of
+// tags will be matched with client queries. If there is a match, the message
+// will be sent to a client.
+func (s *Server) PublishWithTags(msg interface{}, tags map[string]interface{}) error {
+	pubCmd := cmd{op: pub, msg: msg, tags: tags}
+	if s.overflowStrategy == drop {
```
melekes: yes, thanks

pubsub/example_test.go
melekes: no, but I've found that I like the packages providing a clean example of usage. For example https://github.com/go-kit/kit/blob/master/log/example_test.go

pubsub/helpers_test.go
melekes: sure

pubsub/helpers_test.go
melekes: nice idea! actually, found the same function in Tendermint. let's do that

pubsub/helpers_test.go
melekes: ok

pubsub/query/parser_test.go
melekes: I was thinking about adding fuzzy testing (https://github.com/google/gofuzz, https://github.com/dvyukov/go-fuzz). maybe worth doing later

pubsub/query/query.go
melekes: I think so, yeah https://play.golang.org/p/grvSZciTO1

pubsub/example_test.go
melekes: https://golang.org/src/time/example_test.go another example from stdlib

pubsub/query/query_test.go
```diff
@@ -0,0 +1,44 @@
...
+		err     bool
+		matches bool
+	}{
+		{"tm.events.type=NewBlock", map[string]interface{}{"tm.events.type": "NewBlock"}, false, true},
+
+		{"tx.gas > 7", map[string]interface{}{"tx.gas": 8}, false, true},
+		{"body.weight >= 3.5", map[string]interface{}{"body.weight": 3.5}, false, true},
+		{"account.balance < 1000.0", map[string]interface{}{"account.balance": 900}, false, true},
+		{"apples.kg <= 4", map[string]interface{}{"apples.kg": 4.0}, false, true},
+
```
melekes: sure

pubsub/query/query.go
melekes: if u pass something that is not a number (i.e. query = "gas > 7", but tags = { "gas" : "unknown" }. I've changed the code to `panic` in this case and added more type convertions.

pubsub/pubsub_test.go
```diff
@@ -0,0 +1,130 @@
...
+func TestBufferCapacity(t *testing.T) {
+	t.Parallel()
+
+	s := pubsub.NewServer(pubsub.BufferCapacity(2))
+
+	err := s.Publish("Nighthawk")
+	require.NoError(t, err)
+	err = s.Publish("Sage")
+	require.NoError(t, err)
+}
```
melekes: it blocks if `BufferCapacity` is not working

pubsub/pubsub_test.go
```diff
@@ -0,0 +1,130 @@
...
+	defer s.Stop()
+
+	ch := make(chan interface{})
+	s.Subscribe(query.Empty{}, ch)
+	err := s.Publish("Lester")
+	require.NoError(t, err)
+	assertReceive(t, "Lester", ch)
+
+	s.Unsubscribe(ch)
+	close(ch)
```
melekes: yes, good case, thanks

pubsub/query/parser_test.go
ebuchman: cool!

pubsub/example_test.go
ebuchman: neato

pubsub/pubsub.go
```diff
@@ -0,0 +1,208 @@
...
+
+// OverflowStrategyWait will tell the server to block and wait for some time
+// for server to process other messages. Default strategy.
+func OverflowStrategyWait() func(*Server) {
+	return func(s *Server) {
+		s.overflowStrategy = wait
+	}
+}
+
+// Subscribe returns a channel on which messages matching the given query can
```
melekes: No. It used to before. It is a question of control: who should be in control of a channel? and who should decide on the capacity of that channel? subscriber seems to be the more natural choice since only it knows how fast it will process events. 

like subscriber A wants `out` channel to be buffered (5) so it could read 5 events in a row and do something with them. 

Does it make sense?

ps. noticed this pattern here https://blog.golang.org/pipelines

pubsub/pubsub.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func (s *Server) loop(state state) {
+loop:
+	for cmd := range s.cmds {
+		switch cmd.op {
+		case unsub:
+			state.remove(cmd.ch)
+		case shutdown:
+			state.removeAll()
```
melekes: I think it will be garbage collected with the server

pubsub/pubsub.go
melekes: ou. missed that. is there a need? In Tendermint, for example, we usually listen for an event of type X and decode it upon receiving (`Unwrap().(NewBlockData...`). For WS, I created `subscriptions map[query]channel`. https://github.com/tendermint/tendermint/pull/568#discussion_r125291284 and still thinking about creating a `pubsub.Client` to hold this. Or maybe we should go with this multi subscriptions. Will think about it.

pubsub/pubsub.go
```diff
@@ -0,0 +1,208 @@
...
+		// tm.events.type=NewBlock" and tags = {"tm.events.type": "NewBlock"}
+		//
+		// Also, to reduce memory consumption we can store query string <-> Query
+		// pairs in an LRU cache (using reference counter to count number of
+		// clients using this query).
+		if q.Matches(tags) {
+			// non-blocking write
+			select {
+			case ch <- msg:
+			default:
```
melekes: if the subscriber died/stopped OR is not able to keep up with the server, do we want to wait for it in any of these cases? maybe... the same set of strategies will do (wait, wait 10ms, drop)

pubsub/pubsub.go
melekes: https://stackoverflow.com/q/8593645

melekes:
Addressed https://github.com/tendermint/tmlibs/pull/28#discussion_r126158989 and https://github.com/tendermint/tmlibs/pull/28#discussion_r126156643. I think the interface looks nice now. The implementation became a bit more complex though.

melekes:
@ethanfrey @ebuchman could you take a second look at this PR? anyone else is welcomed too of course.

pubsub/pubsub.go
ebuchman: This is fine. I was being silly.

pubsub/pubsub.go
ebuchman: it makes sense, we just need to update the comment

pubsub/pubsub.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+	}
+}
+
+// Returns capacity of the internal server's queue.
+func (s Server) BufferCapacity() int {
+	return s.cmdsCap
+}
+
+// Subscribe returns a channel on which messages matching the given query can
```
ebuchman: Update the comment. It does not return a channel

pubsub/pubsub.go
```diff
@@ -0,0 +1,255 @@
...
+	msg      interface{}
+	tags     map[string]interface{}
+}
+
+// Query defines an interface for a query to be used for subscribing.
+type Query interface {
+	Matches(tags map[string]interface{}) bool
+}
+
+// Server allows clients to subscribe/unsubscribe for messages, pubsling
```
ebuchman: pubsling lol

pubsub/pubsub.go
```diff
@@ -0,0 +1,255 @@
...
+func (s *Server) Unsubscribe(ctx context.Context, clientID string, query Query) error {
+	select {
+	case s.cmds <- cmd{op: unsub, clientID: clientID, query: query}:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+// Unsubscribe unsubscribes the given channel. Blocking.
```
ebuchman: the given client

pubsub/pubsub.go
```diff
@@ -0,0 +1,255 @@
...
+
+	// add client if needed
+	if _, ok := state.clients[clientID]; !ok {
+		state.clients[clientID] = make(map[Query]struct{})
+	}
+	state.clients[clientID][q] = struct{}{}
+
+	// create subscription
+	clientToChannelMap := state.queries[q]
+	clientToChannelMap[clientID] = ch
```
ebuchman: why are we doing this twice? didn't we already do it 10 lines up ?

pubsub/pubsub_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+	ch3 := make(chan interface{}, 1)
+	err = s.Subscribe(ctx, "client-3", query.MustParse("tm.events.type=NewRoundStep AND abci.account.name=Igor AND abci.invoice.number = 10"), ch3)
+	require.NoError(t, err)
+	err = s.PublishWithTags(ctx, "Valeria Richards", map[string]interface{}{"tm.events.type": "NewRoundStep"})
+	require.NoError(t, err)
+	assert.Zero(t, len(ch3))
+}
+
+func TestClientResubscribes(t *testing.T) {
```
ebuchman: Resubscribes? What are we testing here? Are we supposed to unsubscribe and resubscribe ?

pubsub/pubsub_test.go
```diff
@@ -0,0 +1,225 @@
...
+	ch1, ch2 := make(chan interface{}, 1), make(chan interface{}, 1)
+	err := s.Subscribe(ctx, clientID, query.MustParse("tm.events.type=NewBlock"), ch1)
+	require.NoError(t, err)
+	err = s.Subscribe(ctx, clientID, query.MustParse("tm.events.type=NewBlockHeader"), ch2)
+	require.NoError(t, err)
+
+	err = s.UnsubscribeAll(ctx, clientID)
+	require.NoError(t, err)
+
+	err = s.Publish(ctx, "Nick Fury")
```
ebuchman: Is this a proper test? The subscriptions have queries and this has no tags, so shouldn't we hear nothing even if we did not unsubscribe ?

pubsub/pubsub_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+func TestBufferCapacity(t *testing.T) {
+	s := pubsub.NewServer(pubsub.BufferCapacity(2))
+	s.SetLogger(log.TestingLogger())
+
+	ctx := context.Background()
+	err := s.Publish(ctx, "Nighthawk")
+	require.NoError(t, err)
+	err = s.Publish(ctx, "Sage")
+	require.NoError(t, err)
```
ebuchman: Now that we are using context, can we publish a third time and assert it returns an error ?

pubsub/pubsub.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	delete(state.clients, clientID)
+}
+
+func (state *state) send(msg interface{}, tags map[string]interface{}) {
+	for q, clientToChannelMap := range state.queries {
+		if q.Matches(tags) {
+			for _, ch := range clientToChannelMap {
+				ch <- msg
```
ebuchman: blocking on client. is this really what we want? I believe before we did not block and would let it silently fail. 

```
if the subscriber died/stopped OR is not able to keep up with the server, do we want to wait for it in any of these cases? maybe... the same set of strategies will do (wait, wait 10ms, drop)
```

I think for now we should not block and log an error if the send fails. and then we can revisit if we want better strategies here.

pubsub/pubsub_test.go
melekes: in the ideal world, yes (you are supposed to unsubscribe, then subscribe again). nice to have a test just in case

pubsub/pubsub_test.go
melekes: `query.Empty{}` will match with any tags (even empty set). maybe we should remove `Publish` method, not sure...

pubsub/pubsub_test.go
melekes: yes, good idea!

pubsub/pubsub.go
melekes: lol, fixed

pubsub/pubsub.go
melekes: good catch! fixed

pubsub/pubsub.go
melekes: fixed

pubsub/pubsub.go
melekes: fixed

pubsub/pubsub_test.go
ebuchman: I meant this test doesn't seem to actually test that - it doesn't unsubscribe!

melekes: Oh, sorry. wrong naming... it should be `TestClientSubscribesTwice`

pubsub/pubsub_test.go
ebuchman: Not sure I understand why query.Empty{} is relevant here. The subscribtions have queries and the publish has no tags - so even without UnsubscribeAll, we should never receive on the channels. Indeed, if I comment out UnsubscribeAll, the test just hangs.

ebuchman: So I think we can fix this test just be removing the queries from the subscriptions, ie. `err := s.Subscribe(ctx, clientID, query.Empty{}, ch1)`.  Then if we remove UnsubscribeAll in the test, it fails properly. Does that make sense ?

melekes: ou, now I get it. yeah, totally makes sense. stupid me

PR #31, opened 2017-07-21
[pubsub/query] quote values using single quotes

melekes:
This fixes the problem with base-16 encoded values which may start with
digits: 015AB.... In such cases, the parser recognizes them as numbers
but fails to parse because of the follow-up characters (AB).

```
failed to parse tm.events.type=Tx AND hash=136E18F7E4C348B780CF873A0BF43922E5BAFA63:
parse error near digit (line 1 symbol 31 - line 1 symbol 32):
  "6"
```

So, from now on we should quote any values. This seems to be the way
Postgresql has chosen.


ebuchman:
Can we add a test for this?

Another option is to force hex values to be `0x` prefixed

melekes:
> Can we add a test for this?

What do you mean? I've modified all the test cases.

> Another option is to force hex values to be 0x prefixed

Postgresql does quoting, so it probably will be more familiar to regular folks

ebuchman:
> What do you mean? I've modified all the test cases.

I mean a test for the hex case

> Postgresql does quoting, so it probably will be more familiar to regular folks

Ok. I just want us to be aware of what we're doing elsewhere and try to maintain consistency. I think currently it kind of varies depending on what interfacing you're talking to. Maybe @zramsay can help us document how hex/string/int handling works on all the interfaces so we can standardize. I'm happy to follow the lead of an established player like postgres tho :)

melekes:
> I just want us to be aware of what we're doing elsewhere and try to maintain consistency. I think currently it kind of varies depending on what interfacing you're talking to.

agree.

PR #32, opened 2017-07-28
Fix rename /root/.tendermint_test/consensus_replay_test/priv_validator.json.new /root/.tendermint_test/consensus_replay_test/priv_validator.json: no such file or directory

melekes:
See commits description for details.

### DIFF
--- a/common/os.go
+++ b/common/os.go
@@ -93,9 +93,8 @@ func MustWriteFile(filePath string, contents []byte, mode os.FileMode) {
 	}
 }
 
-// Writes to newBytes to filePath.
-// Guaranteed not to lose *both* oldBytes and newBytes,
-// (assuming that the OS is perfect)
+// WriteFileAtomic writes newBytes to temp and atomically moves to filePath
+// when everything else succeeds.
 func WriteFileAtomic(filePath string, newBytes []byte, mode os.FileMode) error {
 	// If a file already exists there, copy to filePath+".bak" (overwrite anything)
 	if _, err := os.Stat(filePath); !os.IsNotExist(err) {
@@ -108,13 +107,27 @@ func WriteFileAtomic(filePath string, newBytes []byte, mode os.FileMode) error {
 			return fmt.Errorf("Could not write file %v. %v", filePath+".bak", err)
 		}
 	}
-	// Write newBytes to filePath.new
-	err := ioutil.WriteFile(filePath+".new", newBytes, mode)
+	f, err := ioutil.TempFile("", "")
 	if err != nil {
-		return fmt.Errorf("Could not write file %v. %v", filePath+".new", err)
+		return err
 	}
-	// Move filePath.new to filePath
-	err = os.Rename(filePath+".new", filePath)
+	_, err = f.Write(newBytes)
+	if err == nil {
+		err = f.Sync()
+	}
+	if closeErr := f.Close(); err == nil {
+		err = closeErr
+	}
+	if permErr := os.Chmod(f.Name(), mode); err == nil {
+		err = permErr
+	}
+	if err == nil {
+		err = os.Rename(f.Name(), filePath)
+	}
+	// any err should result in full cleanup
+	if err != nil {
+		os.Remove(f.Name())
+	}
 	return err
 }
 
--- a/common/os.go
+++ b/common/os.go
@@ -96,17 +96,6 @@ func MustWriteFile(filePath string, contents []byte, mode os.FileMode) {
 // WriteFileAtomic writes newBytes to temp and atomically moves to filePath
 // when everything else succeeds.
 func WriteFileAtomic(filePath string, newBytes []byte, mode os.FileMode) error {
-	// If a file already exists there, copy to filePath+".bak" (overwrite anything)
-	if _, err := os.Stat(filePath); !os.IsNotExist(err) {
-		fileBytes, err := ioutil.ReadFile(filePath)
-		if err != nil {
-			return fmt.Errorf("Could not read file %v. %v", filePath, err)
-		}
-		err = ioutil.WriteFile(filePath+".bak", fileBytes, mode)
-		if err != nil {
-			return fmt.Errorf("Could not write file %v. %v", filePath+".bak", err)
-		}
-	}
 	f, err := ioutil.TempFile("", "")
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


common/os.go
```diff
@@ -93,28 +93,30 @@ func MustWriteFile(filePath string, contents []byte, mode os.FileMode) {
...
-		err = ioutil.WriteFile(filePath+".bak", fileBytes, mode)
-		if err != nil {
-			return fmt.Errorf("Could not write file %v. %v", filePath+".bak", err)
-		}
+	f, err := ioutil.TempFile("", "")
+	if err != nil {
+		return err
+	}
+	_, err = f.Write(newBytes)
+	if err == nil {
```
ebuchman: cool monadic-like error handling. have you seen this pattern elsewhere?

common/os.go
```diff
@@ -93,28 +93,30 @@ func MustWriteFile(filePath string, contents []byte, mode os.FileMode) {
...
-	if _, err := os.Stat(filePath); !os.IsNotExist(err) {
-		fileBytes, err := ioutil.ReadFile(filePath)
-		if err != nil {
-			return fmt.Errorf("Could not read file %v. %v", filePath, err)
-		}
-		err = ioutil.WriteFile(filePath+".bak", fileBytes, mode)
-		if err != nil {
-			return fmt.Errorf("Could not write file %v. %v", filePath+".bak", err)
-		}
+	f, err := ioutil.TempFile("", "")
```
ebuchman: Shouldn't we use something deterministic? Otherwise how will we recover?

common/os.go
```diff
@@ -93,28 +93,30 @@ func MustWriteFile(filePath string, contents []byte, mode os.FileMode) {
...
+		err = f.Sync()
+	}
+	if closeErr := f.Close(); err == nil {
+		err = closeErr
+	}
+	if permErr := os.Chmod(f.Name(), mode); err == nil {
+		err = permErr
+	}
+	if err == nil {
+		err = os.Rename(f.Name(), filePath)
```
ebuchman: Why don't we back up the original file ?

ebuchman:
What prompted this bug? Perhaps we can just make sure the dir exists? 

Regardless, I think we were missing an fsync call!

common/os.go
melekes: Recover from what? FS error? Atomic means we either write to file or not. The issue with the old impl was that **it was not really atomic**. This is why I was getting failures in consensus tests and that is why (my guess) we sometimes see the same error on CI.

common/os.go
melekes: golang source code

common/os.go
melekes: we save priv_validator every sec. So if something bad happens, we will **overwrite old good backup with something bad**. That said, I don't see why we need backups here. Do you?

melekes:
> Regardless, I think we were missing an fsync call!

https://github.com/tendermint/tmlibs/blob/d1ca2c6f838fe7428b7da6f62ebf293d4d91b44b/common/os.go#L105

common/os.go
ebuchman: Yeh I mean if we crash before the rename is done, we won't know where the temp file is. I suppose its not a real issue, just trying to wrap my head around it all. 

Can you clarify how the old impl was not really atomic? Obviously missing fsync, and it didn't clean up. Was that it ?

ebuchman: >  I suppose its not a real issue

Because we could crash right before we write to the temp file too. So we'd be marginally better off if we knew where the temp file was, but not much.

common/os.go
ebuchman: we also use this for the p2p.Addrbook. I don't know enough about renaming. If something goes wrong, wouldn't it be good to have a backup of the original file ?

common/os.go
melekes: goroutine A: write to X.new
goroutine B: write to X.new
goroutine A: rename X.new to X
goroutine B: panic when trying to rename X.new to X

new impl eliminates this error by using temp files

common/os.go
melekes: > If something goes wrong, wouldn't it be good to have a backup of the original file?

1) save interval is too short. 2min. nobody reacts so fast. In DB world there are usually multiple backups (24h, 12h, 6h, 3h, 1h, 30min, latest). latest is used when DB crashed.

2) if something goes wrong and TM continue to work: we write bad things to copy and original file anyway
    if something goes wrong and TM crashes: we just use latest saved changes

common/os.go
ebuchman: Is this what was happening? Multiple go-routines were trying to save the same file?! If so we probably have a deeper issue on our hands

melekes: That was my understanding at the time. If we start 4 consensus state instances, they all will attempt to write to the same file.

ebuchman: Can you point to how this happens? They're supposed to all have independent data dirs, eg. https://github.com/tendermint/tendermint/blob/master/consensus/common_test.go#L344

odeke-em: @ebuchman in regards to https://github.com/tendermint/tmlibs/pull/32/files#r132596644, if we crash/panic, as long as we have a `defer os.RemoveAll(f.Name())` then it'll clean up the temp file. Am assuming that your concern was that we don't know the path/name of the new temp file, but that's in `f.Name()`

common/os.go
```diff
@@ -93,28 +93,30 @@ func MustWriteFile(filePath string, contents []byte, mode os.FileMode) {
...
 	}
-	// Write newBytes to filePath.new
-	err := ioutil.WriteFile(filePath+".new", newBytes, mode)
+	// any err should result in full cleanup
 	if err != nil {
-		return fmt.Errorf("Could not write file %v. %v", filePath+".new", err)
+		os.Remove(f.Name())
```
odeke-em: @melekes IMHO I think making this a defer closer to the `f, err := ioutil.TempFile` call would be useful for readibility and might take away @ebuchman's concern in 	https://github.com/tendermint/tmlibs/pull/32/files#r134935886 so I'd say
```go
defer func() {
      if err != nil {
         os.Remove(f.Name())
      }
}()
```
we use the same `err` variable anyways so in the defer suffices.

Now I see a problem here with leaving the temp directory with undeleted files, since at no point will this file be removed if err == nil, unless  am mistaken.

odeke-em: Hahah in the review conversation this function looks so spread out but the function is actually tiny, please ignore my previous remarks, I hadn't opened the full file to see that the function was tiny :)

PR #33, opened 2017-07-31
http: http-utils added after extraction

odeke-em:
Found common http utils that were being multiply duplicated across
many libraries and since am moving things in basecoin/unstable to
add for more functionality, it's better to put them in one
place.

Utilities and tests added:
- [X] ParseRequestJSON
- [X] ParseAndValidateRequestJSON
- [X] WriteCode
- [X] WriteError
- [X] WriteSuccess
- [X] ErrorResponse

### DIFF
--- a/common/http.go
+++ b/common/http.go
@@ -1,0 +1,153 @@
+package common
+
+import (
+	"encoding/json"
+	"io"
+	"net/http"
+
+	"gopkg.in/go-playground/validator.v9"
+
+	"github.com/pkg/errors"
+)
+
+type ErrorResponse struct {
+	Success bool `json:"success,omitempty"`
+
+	// Err is the error message if Success is false
+	Err string `json:"error,omitempty"`
+
+	// Code is set if Success is false
+	Code int `json:"code,omitempty"`
+}
+
+// ErrorWithCode makes an ErrorResponse with the
+// provided err's Error() content, and status code.
+// It panics if err is nil.
+func ErrorWithCode(err error, code int) *ErrorResponse {
+	return &ErrorResponse{
+		Err:  err.Error(),
+		Code: code,
+	}
+}
+
+// Ensure that ErrorResponse implements error
+var _ error = (*ErrorResponse)(nil)
+
+func (er *ErrorResponse) Error() string {
+	return er.Err
+}
+
+// Ensure that ErrorResponse implements httpCoder
+var _ httpCoder = (*ErrorResponse)(nil)
+
+func (er *ErrorResponse) HTTPCode() int {
+	return er.Code
+}
+
+var errNilBody = errors.Errorf("expecting a non-nil body")
+
+// FparseJSON unmarshals into save, the body of the provided reader.
+// Since it uses json.Unmarshal, save must be of a pointer type
+// or compatible with json.Unmarshal.
+func FparseJSON(r io.Reader, save interface{}) error {
+	if r == nil {
+		return errors.Wrap(errNilBody, "Reader")
+	}
+
+	dec := json.NewDecoder(r)
+	if err := dec.Decode(save); err != nil {
+		return errors.Wrap(err, "Decode/Unmarshal")
+	}
+	return nil
+}
+
+// ParseRequestJSON unmarshals into save, the body of the
+// request. It closes the body of the request after parsing.
+// Since it uses json.Unmarshal, save must be of a pointer type
+// or compatible with json.Unmarshal.
+func ParseRequestJSON(r *http.Request, save interface{}) error {
+	if r == nil || r.Body == nil {
+		return errNilBody
+	}
+	defer r.Body.Close()
+
+	return FparseJSON(r.Body, save)
+}
+
+// ParseRequestAndValidateJSON unmarshals into save, the body of the
+// request and invokes a validator on the saved content. To ensure
+// validation, make sure to set tags "validate" on your struct as
+// per https://godoc.org/gopkg.in/go-playground/validator.v9.
+// It closes the body of the request after parsing.
+// Since it uses json.Unmarshal, save must be of a pointer type
+// or compatible with json.Unmarshal.
+func ParseRequestAndValidateJSON(r *http.Request, save interface{}) error {
+	if r == nil || r.Body == nil {
+		return errNilBody
+	}
+	defer r.Body.Close()
+
+	return FparseAndValidateJSON(r.Body, save)
+}
+
+// FparseAndValidateJSON like FparseJSON unmarshals into save,
+// the body of the provided reader. However, it invokes the validator
+// to check the set validators on your struct fields as per
+// per https://godoc.org/gopkg.in/go-playground/validator.v9.
+// Since it uses json.Unmarshal, save must be of a pointer type
+// or compatible with json.Unmarshal.
+func FparseAndValidateJSON(r io.Reader, save interface{}) error {
+	if err := FparseJSON(r, save); err != nil {
+		return err
+	}
+	return validate(save)
+}
+
+var theValidator = validator.New()
+
+func validate(obj interface{}) error {
+	return errors.Wrap(theValidator.Struct(obj), "Validate")
+}
+
+// WriteSuccess JSON marshals the content provided, to an HTTP
+// response, setting the provided status code and setting header
+// "Content-Type" to "application/json".
+func WriteSuccess(w http.ResponseWriter, data interface{}) {
+	WriteCode(w, data, 200)
+}
+
+// WriteCode JSON marshals content, to an HTTP response,
+// setting the provided status code, and setting header
+// "Content-Type" to "application/json". If JSON marshalling fails
+// with an error, WriteCode instead writes out the error invoking
+// WriteError.
+func WriteCode(w http.ResponseWriter, out interface{}, code int) {
+	blob, err := json.MarshalIndent(out, "", "  ")
+	if err != nil {
+		WriteError(w, err)
+	} else {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(code)
+		w.Write(blob)
+	}
+}
+
+type httpCoder interface {
+	HTTPCode() int
+}
+
+// WriteError is a convenience function to write out an
+// error to an http.ResponseWriter, to send out an error
+// that's structured as JSON i.e the form
+//    {"error": sss, "code": ddd}
+// If err implements the interface HTTPCode() int,
+// it will use that status code otherwise, it will
+// set code to be http.StatusBadRequest
+func WriteError(w http.ResponseWriter, err error) {
+	code := http.StatusBadRequest
+	if httpC, ok := err.(httpCoder); ok {
+		code = httpC.HTTPCode()
+	}
+
+	WriteCode(w, ErrorWithCode(err, code), code)
+}
--- a/common/http_test.go
+++ b/common/http_test.go
@@ -1,0 +1,250 @@
+package common_test
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"reflect"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/tendermint/tmlibs/common"
+)
+
+func TestWriteSuccess(t *testing.T) {
+	w := httptest.NewRecorder()
+	common.WriteSuccess(w, "foo")
+	assert.Equal(t, w.Code, 200, "should get a 200")
+}
+
+var blankErrResponse = new(common.ErrorResponse)
+
+func TestWriteError(t *testing.T) {
+	tests := [...]struct {
+		msg  string
+		code int
+	}{
+		0: {
+			msg:  "this is a message",
+			code: 419,
+		},
+	}
+
+	for i, tt := range tests {
+		w := httptest.NewRecorder()
+		msg := tt.msg
+
+		// First check without a defined code, should send back a 400
+		common.WriteError(w, errors.New(msg))
+		assert.Equal(t, w.Code, http.StatusBadRequest, "#%d: should get a 400", i)
+		blob, err := ioutil.ReadAll(w.Body)
+		if err != nil {
+			assert.Failf(t, "expecting a successful ioutil.ReadAll", "#%d", i)
+			continue
+		}
+
+		recv := new(common.ErrorResponse)
+		if err := json.Unmarshal(blob, recv); err != nil {
+			assert.Failf(t, "expecting a successful json.Unmarshal", "#%d", i)
+			continue
+		}
+
+		assert.Equal(t, reflect.DeepEqual(recv, blankErrResponse), false, "expecting a non-blank error response")
+
+		// Now test with an error that's .HTTPCode() int conforming
+
+		// Reset w
+		w = httptest.NewRecorder()
+
+		common.WriteError(w, common.ErrorWithCode(errors.New("foo"), tt.code))
+		assert.Equal(t, w.Code, tt.code, "case #%d", i)
+	}
+}
+
+type marshalFailer struct{}
+
+var errFooFailed = errors.New("foo failed here")
+
+func (mf *marshalFailer) MarshalJSON() ([]byte, error) {
+	return nil, errFooFailed
+}
+
+func TestWriteCode(t *testing.T) {
+	codes := [...]int{
+		0: http.StatusOK,
+		1: http.StatusBadRequest,
+		2: http.StatusUnauthorized,
+		3: http.StatusInternalServerError,
+	}
+
+	for i, code := range codes {
+		w := httptest.NewRecorder()
+		common.WriteCode(w, "foo", code)
+		assert.Equal(t, w.Code, code, "#%d", i)
+
+		// Then for the failed JSON marshaling
+		w = httptest.NewRecorder()
+		common.WriteCode(w, &marshalFailer{}, code)
+		wantCode := http.StatusBadRequest
+		assert.Equal(t, w.Code, wantCode, "#%d", i)
+		assert.True(t, strings.Contains(string(w.Body.Bytes()), errFooFailed.Error()),
+			"#%d: expected %q in the error message", i, errFooFailed)
+	}
+}
+
+type saver struct {
+	Foo int    `json:"foo" validate:"min=10"`
+	Bar string `json:"bar"`
+}
+
+type rcloser struct {
+	closeOnce sync.Once
+	body      *bytes.Buffer
+	closeChan chan bool
+}
+
+var errAlreadyClosed = errors.New("already closed")
+
+func (rc *rcloser) Close() error {
+	var err = errAlreadyClosed
+	rc.closeOnce.Do(func() {
+		err = nil
+		rc.closeChan <- true
+		close(rc.closeChan)
+	})
+	return err
+}
+
+func (rc *rcloser) Read(b []byte) (int, error) {
+	return rc.body.Read(b)
+}
+
+var _ io.ReadCloser = (*rcloser)(nil)
+
+func makeReq(strBody string) (*http.Request, <-chan bool) {
+	closeChan := make(chan bool, 1)
+	buf := new(bytes.Buffer)
+	buf.Write([]byte(strBody))
+	req := &http.Request{
+		Header: make(http.Header),
+		Body:   &rcloser{body: buf, closeChan: closeChan},
+	}
+	return req, closeChan
+}
+
+func TestParseRequestJSON(t *testing.T) {
+	tests := [...]struct {
+		body    string
+		wantErr bool
+		useNil  bool
+	}{
+		0: {wantErr: true, body: ``},
+		1: {body: `{}`},
+		2: {body: `{"foo": 2}`}, // Not that the validate tags don't matter here since we are just parsing
+		3: {body: `{"foo": "abcd"}`, wantErr: true},
+		4: {useNil: true, wantErr: true},
+	}
+
+	for i, tt := range tests {
+		req, closeChan := makeReq(tt.body)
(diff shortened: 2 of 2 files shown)


common/http.go
```diff
@@ -0,0 +1,79 @@
...
+	Code int `json:"code,omitempty"`
+}
+
+var errNilBody = errors.Errorf("expecting a non-nil body")
+
+// ParseRequestJSON unmarshals into save, the body of the
+// request. It closes the body of the request after parsing.
+// Since it uses json.Unmarshal, save must be of a pointer type
+// or compatible with json.Unmarshal.
+func ParseRequestJSON(r *http.Request, save interface{}) error {
```
ethanfrey: Please replace *http.Request with io.ReadCloser and r.Body with r

Makes it more reusable

common/http.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return errors.Wrap(err, "Read Request")
+	}
+	if err := json.Unmarshal(slurp, save); err != nil {
+		return errors.Wrap(err, "Parse")
+	}
+	return nil
+}
+
+func ParseAndValidateRequestJSON(r *http.Request, save interface{}) error {
```
ethanfrey: Same as above

common/http.go
```diff
@@ -0,0 +1,79 @@
...
+	} else {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(code)
+		w.Write(blob)
+	}
+}
+
+func WriteError(w http.ResponseWriter, err error) {
+	resp := &ErrorResponse{
+		Code:  http.StatusBadRequest,
```
ethanfrey: We do want a way to set this, right? 

How about allowing some special interface with a method `HTTPCode() int` and if err implements that, use it, otherwise BadRequest

ethanfrey: Oh, and a helper method/struct to add this info

ethanfrey: Or just a third argument to the function which is probably easier

common/http.go
ethanfrey: I have cases where I would like to parse the Body two times into two different structs. Like first one set of keys, and use those to select the other set I need.

common/http_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"testing"
+
+	"github.com/tendermint/tmlibs/common"
+)
+
+func TestWriteSuccess(t *testing.T) {
+	w := httptest.NewRecorder()
+	common.WriteSuccess(w, "foo")
+	got, want := w.Code, 200
+	if got != want {
```
ethanfrey: We have decided to use testify for all test cases. Can you port this... And the below checks

assert.Equal(want,got) instead of the if. 

Use require instead of assert if you want t.Fatal

common/http_test.go
```diff
@@ -0,0 +1,169 @@
...
+	buf.Write([]byte(strBody))
+	req := &http.Request{
+		Header: make(http.Header),
+		Body:   &rcloser{body: buf, closeChan: closeChan},
+	}
+	return req, closeChan
+}
+
+func TestParseRequestJSON(t *testing.T) {
+	tests := [...]struct {
```
ethanfrey: Why not just use a slice here? Please enlighten me

common/http.go
ethanfrey: Or maybe leave as is. And add another function that just reads the bytes safely. Anyway, i do want to parse that data twice sometimes

odeke-em: Thank you for the request for updates. Sure, I can change it and that might be a less broader API for the io.Reader instead of a io.ReadCloser or *http.Request. Coming up in a few.

common/http.go
odeke-em: Roger that, thanks.

common/http_test.go
odeke-em: Good question. So using a slice is usually nice, but doesn't allow you to seek back to the index easily in case of any error, for example
given an error
```shell
sample_test.go:174 "#5: unknown string"
```

## Slice
```go
tests := []string{"a","b","c","d","e","f","g"}
```

## Indexed slice
```
tests := [...]string{0:"a", 1:"b", 2:"c", 3:"d", 4:"e", 5:"f", 6:"g"}
```
in the case of the slice, you'll have to manually go find the element then edit it, and in many cases since we are humans, might change the wrong element(testing becomes harder to fix)
in the case of indexed slice, you just search /5: and that's it. It's a personal style that I've adopted from dealing with many confusing tests cases.

common/http.go
odeke-em: Yeah before, was operating on the premise that all the users just dumped out an http.StatusBadRequest, but sure great idea to make it changeable. I'll make it so that we type assert to check if it implements the HTTPCode interface then switch up the status code

common/http.go
odeke-em: Done, thanks.

odeke-em:
@ethanfrey thanks for the first round, I've updated the PR as per your comments, PTAL.

common/http_test.go
ethanfrey: Interesting

ethanfrey:
FYI: Please don't rebase during a review (until it is approved).  This makes it much harder for me to track what is changed.  I know this is common practice and we had this discussion recently internally as well.

It's nice to see a few changes and track the evolution.

common/http.go
```diff
@@ -0,0 +1,131 @@
...
+
+// FparseJSON unmarshals into save, the body of the provided reader.
+// Since it uses json.Unmarshal, save must be of a pointer type
+// or compatible with json.Unmarshal.
+func FparseJSON(r io.Reader, save interface{}) error {
+	if r == nil {
+		return errNilBody
+	}
+
+	slurp, err := ioutil.ReadAll(r)
```
ethanfrey: Actually now that I think of it... ` json.NewDecoder()` will give you a Decoder that can parse a stream rather than loading all bytes first.

common/http.go
```diff
@@ -0,0 +1,131 @@
...
+	// Code is set if Success is false
+	Code int `json:"code,omitempty"`
+}
+
+var errNilBody = errors.Errorf("expecting a non-nil body")
+
+// FparseJSON unmarshals into save, the body of the provided reader.
+// Since it uses json.Unmarshal, save must be of a pointer type
+// or compatible with json.Unmarshal.
+func FparseJSON(r io.Reader, save interface{}) error {
```
ethanfrey: Nice job with the more generic Fparse functions and keeping the more generic Parse... from request wrappers.

common/http.go
```diff
@@ -0,0 +1,131 @@
...
+// WriteError is a convenience function to write out an
+// error to an http.ResponseWriter, to send out an error
+// that's structured as JSON i.e the form
+//    {"error": sss, "code": ddd}
+// If err implements the interface HTTPCode() int,
+// it will use that status code otherwise, it will
+// set code to be http.StatusBadRequest
+func WriteError(w http.ResponseWriter, err error) {
+	code := http.StatusBadRequest
+	if httpC, ok := err.(httpCoder); ok {
```
ethanfrey: Nice, more flexible.

common/http.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		WriteError(w, err)
+	} else {
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(code)
+		w.Write(blob)
+	}
+}
+
+type httpCoder interface {
```
ethanfrey: Can you add a simple `WithCode(error, int) error` function that wraps the error and adds a HTTPCode method to it?  It would make this easier to use....

`WriteError(w, WithCode(err, 406))`

common/http_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+var blankErrResponse = new(common.ErrorResponse)
+
+func TestWriteError(t *testing.T) {
+	w := httptest.NewRecorder()
+	msg := "this is an error"
+	common.WriteError(w, errors.New(msg))
+	assert.Equal(t, w.Code, http.StatusBadRequest, "should get a 400")
+	blob, err := ioutil.ReadAll(w.Body)
+	require.Equal(t, err, nil, "expecting a successful ioutil.ReadAll")
```
ethanfrey: There is also `require.Nil`

common/http_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+func TestWriteSuccess(t *testing.T) {
+	w := httptest.NewRecorder()
+	common.WriteSuccess(w, "foo")
+	assert.Equal(t, w.Code, 200, "should get a 200")
+}
+
+var blankErrResponse = new(common.ErrorResponse)
+
+func TestWriteError(t *testing.T) {
```
ethanfrey: Please add a test here that checks what happens when it implements httpCoder

common/http_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/tendermint/tmlibs/common"
+)
+
+func TestWriteSuccess(t *testing.T) {
```
ethanfrey: also interesting to see what happens when the object returns an error on json.Marshal, that the error code is set properly.

common/http.go
```diff
@@ -0,0 +1,131 @@
...
+	"io/ioutil"
+	"net/http"
+
+	"gopkg.in/go-playground/validator.v9"
+
+	"github.com/pkg/errors"
+)
+
+type ErrorResponse struct {
+	Success bool `json:"success,omitempty"`
```
ethanfrey: Since `success` is always set to `false`, `omitempty` will make it disappear. 
Maybe we should remove it, but then from the code below as well.

common/http.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+var errNilBody = errors.Errorf("expecting a non-nil body")
+
+// FparseJSON unmarshals into save, the body of the provided reader.
+// Since it uses json.Unmarshal, save must be of a pointer type
+// or compatible with json.Unmarshal.
+func FparseJSON(r io.Reader, save interface{}) error {
+	if r == nil {
+		return errNilBody
```
ethanfrey: Noted other errors are `errors.Wrap`d but this not.  Is there a reason?

common/http_test.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	return req, closeChan
+}
+
+func TestParseRequestJSON(t *testing.T) {
+	tests := [...]struct {
+		body    string
+		want    interface{}
+		wantErr bool
+	}{
```
ethanfrey: want is always nil? shouldn't you provide a `new(InputStruct)` or something like that.  Otherwise, of course they all error. It would be good to see some positive results as well.

common/http_test.go
```diff
@@ -0,0 +1,171 @@
...
+	require.Equal(t, reflect.DeepEqual(sav, &saver{Foo: 100, Bar: "blockchain"}), true, "should have parsed both")
+}
+
+func TestParseAndValidateRequestJSON(t *testing.T) {
+	tests := [...]struct {
+		body    string
+		want    interface{}
+		wantErr bool
+	}{
+		0: {wantErr: true, body: ``},
```
ethanfrey: Same comment as above: 
want is always nil? shouldn't you provide a `new(saver)` or something like that.  Otherwise, of course they all error. It would be good to see some positive results as well.  Feel these tests were 80% finished.

common/http.go
odeke-em: So ErrorResponse.Success is used in tendermint/basecoin/cmd/baseserver and tendermint/go-crypto/keys so IMO we still need it.

common/http.go
odeke-em: Not really, just a migration without too much thought plus the other errors are wrapped because we received an error and wanted to add context; here `r == nil` as opposed to `err != nil` but the two cases could be made into the same thing; Seems like you are leaning towards errors.Wrap right?

common/http.go
odeke-em: Good train of thought, I had briefly considered that but thought ioutil.ReadAll sufficed. I'll update it to use json.NewDecoder, thanks.

common/http_test.go
odeke-em: No, want is never used in the test, it was just an artifact of a changed test, I'll remove it.

common/http_test.go
odeke-em: No, want is never used in the test, it was just an artifact of a changed test, I'll remove it.
And there are some positive results, please see the usage of wantErr:true and not, that tests the different scenarios.

common/http_test.go
odeke-em: Yeah as mentioned in https://github.com/tendermint/tmlibs/pull/33#discussion_r130777591, there are also positive results which is what `wantErr: true` helps define the wanted errors from non wanted errors, please see the test body in the for loop.

common/http.go
odeke-em: In particular it is used in tendermint/client/rest/handlers.go#DeleteKey to indicate the response to the caller, on the wire.

common/http_test.go
odeke-em: Nice, thanks, I'll use that.

odeke-em:
Thank you @ethanfrey for the second round of review, please see the latest PR. Oh so it is the style at Tendermint to keep all revisions of updates in history? My bad I didn't know about that and was always rebasing all my commits so that I'd have a single commit at the end so anyways now I won't anymore.

common/http.go
ethanfrey: I like the FparseRequest format.

ethanfrey:
Rebasing at the end is not bad.  I usually rebase together minor fixups, bugfixes, but allow multiple commits per pr.  If they are adding separate pieces  of code (our prs are often rather large, >500 lines and touch a few different aspects), it makes it easier to follow. This PR would make sense as one commit though, as it is just different revisions of the same file.

The point was only to wait until it was all approved, then collapse commits and rebase on the latest develop to get a clean history.

common/http.go
ethanfrey: ugh, i was abusing that type, huh?  okay, leave it, but maybe remove the `omitempty`

common/http.go
```diff
@@ -14,29 +13,50 @@ import (
 type ErrorResponse struct {
 	Success bool `json:"success,omitempty"`
 
-	// Error is the error message if Success is false
-	Error string `json:"error,omitempty"`
+	// Err is the error message if Success is false
+	Err string `json:"error,omitempty"`
```
ethanfrey: Yeah, I think that is a nicer name

common/http.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// WriteError is a convenience function to write out an
+// error to an http.ResponseWriter, to send out an error
+// that's structured as JSON i.e the form
+//    {"error": sss, "code": ddd}
+// If err implements the interface HTTPCode() int,
+// it will use that status code otherwise, it will
+// set code to be http.StatusBadRequest
+func WriteError(w http.ResponseWriter, err error) {
```
ethanfrey: I like it very much now, with the ErrorWithCode reuse.

common/http.go
```diff
@@ -0,0 +1,153 @@
...
+	Err string `json:"error,omitempty"`
+
+	// Code is set if Success is false
+	Code int `json:"code,omitempty"`
+}
+
+// ErrorWithCode makes an ErrorResponse with the
+// provided err's Error() content, and status code.
+// It panics if err is nil.
+func ErrorWithCode(err error, code int) *ErrorResponse {
```
ethanfrey: Crazy, so this is both a struct to serialize, and an enhanced error to add the http code in WriteError?

Minimal API and I think this is a good design, I just had to scroll up again to check my eyes.

common/http_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+		assert.Equal(t, reflect.DeepEqual(recv, blankErrResponse), false, "expecting a non-blank error response")
+
+		// Now test with an error that's .HTTPCode() int conforming
+
+		// Reset w
+		w = httptest.NewRecorder()
+
+		common.WriteError(w, common.ErrorWithCode(errors.New("foo"), tt.code))
+		assert.Equal(t, w.Code, tt.code, fmt.Sprintf("#%d: got=%d want=%d", i, w.Code, tt.code))
```
ethanfrey: Good test, but you can write the check shorter.
1) assert.Equal always prints the two values if they don't equal, as a default.
2) The arguments after the comparison values are interpreter by fmt.Sprintf already

This is effectively the same:
`assert.Equal(t, w.Code, tt.code, "#%d", i)`

It will `fmt.Sprintf("#%d", i)` as the message for you, no need to write it out.
And want and got are always in the top of the error message.

common/http_test.go
```diff
@@ -0,0 +1,245 @@
...
+		w := httptest.NewRecorder()
+		common.WriteCode(w, "foo", code)
+		assert.Equal(t, w.Code, code, fmt.Sprintf("#%d", i))
+
+		// Then for the failed JSON marshaling
+		w = httptest.NewRecorder()
+		common.WriteCode(w, &marshalFailer{}, code)
+		wantCode := http.StatusBadRequest
+		assert.Equal(t, w.Code, wantCode, fmt.Sprintf("#%d", i))
+		assert.Equal(t, strings.Contains(string(w.Body.Bytes()), errFooFailed.Error()), true,
```
ethanfrey: There is also `assert.True` for this case.

common/http_test.go
```diff
@@ -0,0 +1,245 @@
...
+	for i, code := range codes {
+		w := httptest.NewRecorder()
+		common.WriteCode(w, "foo", code)
+		assert.Equal(t, w.Code, code, fmt.Sprintf("#%d", i))
+
+		// Then for the failed JSON marshaling
+		w = httptest.NewRecorder()
+		common.WriteCode(w, &marshalFailer{}, code)
+		wantCode := http.StatusBadRequest
+		assert.Equal(t, w.Code, wantCode, fmt.Sprintf("#%d", i))
```
ethanfrey: Nitpick: `assert.Equal(t, w.Code, wantCode, "#%d", i)`

common/http_test.go
```diff
@@ -0,0 +1,245 @@
...
+	return req, closeChan
+}
+
+func TestParseRequestJSON(t *testing.T) {
+	tests := [...]struct {
+		body    string
+		wantErr bool
+		useNil  bool
+	}{
+		0: {wantErr: true, body: ``},
```
ethanfrey: look good the cases

common/http_test.go
```diff
@@ -0,0 +1,245 @@
...
+	require.NotEqual(t, nil, common.FparseJSON(nil, sav), "expecting a nil error report")
+}
+
+func TestParseAndValidateRequestJSON(t *testing.T) {
+	tests := [...]struct {
+		body    string
+		wantErr bool
+		useNil  bool
+	}{
+		0: {wantErr: true, body: ``},
```
ethanfrey: please add at least one test that passes, eg. `{"foo": 20}` or there is no proof that anything will validate successfully.

common/http_test.go
```diff
@@ -0,0 +1,245 @@
...
+		if tt.useNil {
+			req.Body = nil
+		}
+		sav := new(saver)
+		err := common.ParseRequestJSON(req, sav)
+		if tt.wantErr {
+			assert.NotEqual(t, err, nil, fmt.Sprintf("#%d: want non-nil error", i))
+			continue
+		}
+		assert.Equal(t, err, nil, fmt.Sprintf("#%d: want nil error", i))
```
ethanfrey: We never check the value was stored properly on success.... maybe nitpicky but at least one check on foo.

Also, simplifying the assert checks... shorter and I get the full stack trace if there is an error.
`assert.Nil(t, err, "#%d: %+v", i, err)`

ethanfrey:
Okay, I am happy with how it has shaped up.  Nice code I am happy to use :)

Some minor feedback to add a check or two in the Parse tests, and to simplify the assert calls.  I have basically three patterns:

`assert.Equal(expected, got, "#%d", i)`
`assert.Nil(err, "#%d: %+v", i, err)`
`assertTrue(check(), "#%d", i)` (or maybe a little more data about what data you were checking)

They let you know the reason (a is not b, err is not nil, ...) and no need for fmt.Sprintf, it is used by the library internally.

If you do a little cleanup, you can squash the commits and rebase on the latest unstable, and I will merge it in. (I'll be online til like noon your time today).

common/http_test.go
odeke-em: Oops, thought I had one before, but not anymore.

common/http_test.go
odeke-em: Nice, thanks for letting know. Yeah:
1) The signature for assert.Equal kinda tripped me out and I couldn't intuitively guess what it did
http://godoc.org/github.com/stretchr/testify/assert#Equal
screen shot 2017-08-02 at 11 23 36 am

VS

https://golang.org/pkg/fmt/#Printf
screen shot 2017-08-02 at 11 27 04 am

Notice the lack of: `format string, args ...interface{}` which in the case of assert.Equal are `msgAndArgs ...interface{}`

I'll update that in a sec.

common/http_test.go
odeke-em: Done.

common/http_test.go
odeke-em: Done, thanks.

odeke-em: That's done in TestWriteCode, thanks.

common/http.go
odeke-em: Implemented with `ErrorWithCode`

common/http.go
odeke-em: So I am keeping the omitempty so that in cases where we do
`success` and leave the rest out, only `success` is sent on the wire and the opposite, when `err` and `code` are defined, we only send `err` and `code` on the wire.

odeke-em:
Thank you @ethanfrey for the other round of review, I've updated the PR with your comments(thank you), and also squashed my commits into 1. PTAL.

common/http.go
ethanfrey: okay

PR #34, opened 2017-08-04
common/IsDirEmpty: do not mask non-existance errors

odeke-em:
Currently IsDirEmpty returns true, err if it encounters
any error after trying to os.Open the directory.
I noticed this while studying the code and recalled a bug
from an earlier project in which doing the exact same thing
on code without permissions would trip out and falsely report
that the directory was empty.
Given demo.go in https://play.golang.org/p/vhTPU2RiCJ

* Demo:
```shell
$ mkdir -p sample-demo/1 && touch sample-demo/2
$ echo "1st round" && go run demo.go sample-demo
$ sudo chown root sample-demo && sudo chmod 0700 sample-demo
$ echo "2nd round" && go run demo.go sample-demo
```

That then prints out
```shell
1st round
original:: empty: false err: <nil>
updated::  empty: false err: <nil>
2nd round
original:: empty: true err: open data/: permission denied
updated::  empty: false err: open data/: permission denied
```

where in "2nd round", the original code falsely reports that
the directory is empty but that's a permission error.

I could write a code test for it, but that test requires me to change
users and switch to root as a Go user so no point in complicating our
tests, but otherwise it is a 1-to-1 translation between shell and Go.

### DIFF
--- a/common/os.go
+++ b/common/os.go
@@ -48,7 +48,12 @@ func EnsureDir(dir string, mode os.FileMode) error {
 func IsDirEmpty(name string) (bool, error) {
 	f, err := os.Open(name)
 	if err != nil {
-		return true, err //folder is non-existent
+		if os.IsNotExist(err) {
+			return true, err
+		}
+		// Otherwise perhaps a permission
+		// error or some other error.
+		return false, err
 	}
 	defer f.Close()
 


ebuchman:
Interesting. I actually can't find a case where we check the error. 

Do we think it makes sense to say "the dir is empty" if it doesn't even exist? I guess so long as we're still returning an error. But it means consumers should definitely be checking that error. 

Maybe we should remove the function all together and replace it with something more suited to what we use it for.  I can only find it being used in basecoin and merkleeyes and for the same thing both times - to see if the database has been initialized!

PR #35, opened 2017-08-13
common: implement ParseGenesisOrDefault

odeke-em:
For https://github.com/tendermint/ethermint/issues/244.

ParseGenesisOrDefault replaces the step in which while
init-ing ethermint, we have to
* cd $GOPATH/src/github.com/tendermint/ethermint
* ethermint --datadir ~/.ethermint init setup/genesis.json
so we have the serialized record of the setup/genesis.json
that will be used by default, if the file:
a) doesn't exist
b) doesn't have content ie less than {}

+ Why here:
Since $GOPATH/src/tendermint/ethermint/cmd/ethermint
has a bunch of initializers in:
ethUtils "github.com/ethereum/go-ethereum/cmd/utils"
```go
app = ethUtils.NewApp(version.Version, "the ethermint command line interface")
```
whose definition in flags.go
```go
func NewApp(gitcommit, usage string) *cli.App {
  app := cli.NewApp()
  app.Name = filepath.Base(os.Args[0])
```
that depend on os.Args[0]

so we can't be running tests in there without periliously
trying to change os.Args that isn't populated when we run
```shell
$ go test -run=TestParse
```
or in our CI
thus extract and move that utility to here, the common utilities area


ebuchman:
Thanks Emmanuel, but I don't think this is the right place for this - we can't have tmlibs/common importing anything like go-ethereum. rather - let's refactor whatever we need to on ethermint side to put this there.

odeke-em:
Thanks for the feedback @ebuchman, good point there, I see it now, tmlibs/common shouldn't import go-ethereum, that'll bloat the dependency list for things that aren't even related to go-ethereum. Thanks for advocating for that change!

PR #47, opened 2017-09-06
cli: add VerifyFlag, used by Tendermint

odeke-em:
For https://github.com/tendermint/tendermint/pull/636
Updates https://github.com/tendermint/tendermint/issues/628

Add global configuration flag `--verify` which is
used currently, primarily by Tendermint. Since
`cli/setup.go` has the definitions for Tendermint's
global configuration flags, place this one there too.


ebuchman:
Since this is really just helpful for devs in Tendermint, let's not add it here just yet. We should be able to do it as a hidden option directly on the Tendermint command.

odeke-em:
Aye aye, gotcha. SGTM.

Issue #50, opened 2017-09-18
lint: logger should be lint free


zramsay:
@zramsay commented on [Wed Sep 06 2017](https://github.com/tendermint/tendermint/issues/645)

running `gometalinter --vendor --deadline=600s --disable-all --enable=errcheck ./...` throws many errors for the logger. I don't think it makes sense to check all them but rather have the logger not throw an error.

---

@odeke-em commented on [Wed Sep 06 2017](https://github.com/tendermint/tendermint/issues/645#issuecomment-327630112)

@zramsay I've retitled the issue, please feel free to revert it if I was off.

However, I can't reproduce this issue on the latest on `develop` 54c63726b03cbba067d10fa6b69b885c85678d17
<img width="754" alt="screen shot 2017-09-06 at 4 30 48 pm" src="https://user-images.githubusercontent.com/4898263/30137485-c48bddd0-9320-11e7-8799-b5a5bd4ba68d.png">

---

@odeke-em commented on [Wed Sep 06 2017](https://github.com/tendermint/tendermint/issues/645#issuecomment-327630464)

Also, which logger? I can't find a repository called `logger` or a variant of that in tendermint, might we be talking about `tmlibs/logger` instead?

---

@ebuchman commented on [Mon Sep 11 2017](https://github.com/tendermint/tendermint/issues/645#issuecomment-328607652)

@melekes :P 

---

@melekes commented on [Wed Sep 13 2017](https://github.com/tendermint/tendermint/issues/645#issuecomment-329211343)

Yeah, let's change the interface and implementation to not return an error.

PR #51, opened 2017-09-19
logger no longer returns an error

melekes:
See #50


zramsay:
this works when pinned in glide.lock on tendermint. I also ran `make test` and it passed.

CHANGELOG.md
```diff
@@ -1,18 +1,24 @@
 # Changelog
 
+## 0.4.0 (TBA)
+
+BREAKING CHANGES:
+
+- [log] logger functions no longer returns an error
```
ebuchman: let's just add this to 0.3.0

ebuchman: i got it :)

ebuchman:
merged

odeke-em:
Oops, I am too late to the wedding and passed the with the "hold your silence forever" threshold.

I realize that our goal is to pass the linter and that we never use the previously returned errors in our code. However, aren't we losing the functionality of getting back errors e.g when the logger is a distributed logger, for example using bugsnag or uber/go-zap or does tracing and tries to post the data, and the HTTP call fails? Or is this a conversation we'd rather have later on when we need that functionality?

ebuchman:
That's a good conversation to have. But are we really going to check logging errors everywhere, or what ?

melekes:
> when the logger is a distributed logger

@odeke-em and when it is gonna happen?

odeke-em:
Good questions y'all.
`But are we really going to check logging errors everywhere, or what ?`
@ebuchman not really, and I mentioned that we aren't checking for any errors. However, some signatures like https://golang.org/pkg/fmt/#Printf return errors but no one every checks them, doesn't mean we should discard them entirely. On the other hand, I realize that we are being pragmatic here, so in deed if the need for catching errors then we'll have the conversation.

```
@odeke-em and when it is gonna happen?
```
@melekes I am not sure when, I was giving an example, and I assume that given that we are launching Cosmos in December 2017, perhaps people might want to have logs churned out say to Google Cloud logging or Stackdriver. However, no one is doing that right now though.

melekes:
Good example with `fmt` package. Ok, so let's suppose somebody does request a possibility to log to Stackdriver. Will we be forced to go through all of our code and add error checking for all `Log` calls then?

odeke-em:
So it seems to me like if they need to use a tmlibs.Logger, they'll still need to conform to the new signatures but can entirely make their own wrapper and do the error handling on their own (please note that these aren't fully thought out examples)
```go
type logToStackDriver struct {
   stLogger *stackdriver.Logger
   errHandler ErrHandler
}

var _ tmlibs.Logger = (*logToStackDriver)(nil)
func (ltsd *logToStackDriver) Debug(msg string, keyvals ...interface{}) {
    if err := ltsd.stLogger.Debug(msg, keyvals...); err != nil {
       ltsd.errHandler("debug", err)
    }
} 

func (ltsd *logToStackDriver) Info(msg string, keyvals ...interface{}) {
    if err := ltsd.stLogger.Info(msg, keyvals...); err != nil {
       ltsd.errHandler("info", err)
    }
} 

func (ltsd *logToStackDriver) Error(msg string, keyvals ...interface{}) {
    if err := ltsd.stLogger.Error(msg, keyvals...); err != nil {
       ltsd.errHandler("error", err)
    }
} 
```

melekes:
They can write their own driver, for sure. But how do they enable it (replace current logger)?? By forking tendermint?

odeke-em:
@melekes fortunately we made tmlibs.Logger an interface instead of a concrete type, so notice in the code snippet that I put up in https://github.com/tendermint/tmlibs/pull/51#issuecomment-330953559, that logger can now be used in places where tmlibs.Logger is required. I hope I am answering your question.

melekes:
OK. here is the thing: you are talking about people using `log` package and building new drivers on top of it. But it never was my intent to build log package in such a way. log package was created exclusively for Tendermint. If people will want to forward their logs to some external storage (say Logstash), I would suggest to them to redirect Tendermint output to some log forwarder (i.e. `tendermint node > XXXX`. Does it make sense to you?

PR #53, opened 2017-09-22
add metalinter to CI and address some lint warnings

zramsay:


autofile/autofile_test.go
```diff
@@ -1,12 +1,14 @@
+// nolint: goimports
```
ebuchman: why? we should replace the `.` with `cmn`

common/os.go
```diff
@@ -17,7 +18,7 @@ var (
 func TrapSignal(cb func()) {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt)
-	signal.Notify(c, os.Kill)
+	signal.Notify(c, syscall.SIGTERM)
```
ebuchman: this is recomended by the linter? any idea why? os.Kill is just an alias for syscall. SIGTERM, ie. https://golang.org/pkg/os/#Signal

common/service.go
```diff
@@ -151,7 +151,7 @@ func (bs *BaseService) Reset() (bool, error) {
 		return false, nil
 	}
 	// never happens
-	return false, nil
+	return false, nil // nolint: vet
```
ebuchman: whats the problem here?

merkle/simple_tree.go
```diff
@@ -22,6 +22,7 @@ For larger datasets, use IAVLTree.
 
 */
 
+// nolint: goimports
```
ebuchman: same comment

pubsub/query/query.peg.go
```diff
@@ -1,3 +1,4 @@
+// nolint
```
ebuchman: lol, this is auto generated - maybe you should go run the linter on their repo :P

autofile/autofile_test.go
zramsay: this file is not imported anywhere, so goimports throws

common/os.go
zramsay: the error is "cannot trap ..." something something something. I made this fix in tendermint core awhile back and in various places in ipfs

merkle/simple_tree.go
zramsay: file not imported anywehre

common/service.go
zramsay: see comment on L153 ... never happens, so go vet picks up on it

common/os.go
zramsay: https://github.com/ipfs/go-ipfs/pull/3949/files#r119514774

ebuchman: huh - I think this is actually a bug in go-vet than. os.Kill can be trapped - it's literally just an alias for syscall.SIGTERM. Maybe you want to alert them :)

autofile/autofile_test.go
ebuchman: huh? it's a test, why would it be imported anywhere?

merkle/simple_tree.go
ebuchman: I don't understand what that means. its a pkg for external use.

autofile/autofile_test.go
zramsay: it wouldn't but i guess goimports isn't that smart

pubsub/query/query.peg.go
zramsay: meh

merkle/simple_tree.go
zramsay: so leave the nolint and move on?

common/os.go
zramsay: reverted

merkle/simple_tree.go
ebuchman: Are you suggesting I allow lines of code in my codebase that I don't understand?

common/service.go
odeke-em: @zramsay perhaps instead of adding the nolint clause we can actually fix the code
```diff
diff --git a/common/service.go b/common/service.go
index 71fc03c..dc321e5 100644
--- a/common/service.go
+++ b/common/service.go
@@ -146,11 +146,8 @@ func (bs *BaseService) Reset() (bool, error) {
 
 		bs.Quit = make(chan struct{})
 		return true, bs.impl.OnReset()
-	} else {
-		bs.Logger.Debug(Fmt("Can't reset %v. Not stopped", bs.name), "impl", bs.impl)
-		return false, nil
 	}
-	// never happens
+	bs.Logger.Debug(Fmt("Can't reset %v. Not stopped", bs.name), "impl", bs.impl)
 	return false, nil
 }
 ```

or for easier readability
```diff
diff --git a/common/service.go b/common/service.go
index 71fc03c..2d86baa 100644
--- a/common/service.go
+++ b/common/service.go
@@ -140,18 +140,16 @@ func (bs *BaseService) OnStop() {}
 
 // Implements Service
 func (bs *BaseService) Reset() (bool, error) {
-	if atomic.CompareAndSwapUint32(&bs.stopped, 1, 0) {
-		// whether or not we've started, we can reset
-		atomic.CompareAndSwapUint32(&bs.started, 1, 0)
-
-		bs.Quit = make(chan struct{})
-		return true, bs.impl.OnReset()
-	} else {
+	if stopped := atomic.CompareAndSwapUint32(&bs.stopped, 1, 0); !stopped {
 		bs.Logger.Debug(Fmt("Can't reset %v. Not stopped", bs.name), "impl", bs.impl)
 		return false, nil
 	}
-	// never happens
-	return false, nil
+
+	// whether or not we've started, we can reset
+	atomic.CompareAndSwapUint32(&bs.started, 1, 0)
+
+	bs.Quit = make(chan struct{})
+	return true, bs.impl.OnReset()
 }
 
 // Implements Service
```

common/os.go
```diff
@@ -18,7 +17,7 @@ var (
 func TrapSignal(cb func()) {
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt)
-	signal.Notify(c, syscall.SIGTERM)
+	signal.Notify(c, os.Kill) // nolint: megacheck
```
ebuchman: turns out we do need to change this - os.Kill can't be caught!

https://github.com/dominikh/go-staticcheck/issues/26

Issue #55, opened 2017-09-25
db: MemDB's Close code is invalid


odeke-em:
Currently db.MemDB's Close code is
https://github.com/tendermint/tmlibs/blob/9997e3a3b46db1d2f88aa9816ed0e7915dad6ac1/db/mem_db.go#L54-L58

The bug here is us mistakenly setting db to nil expecting the pointer of db to be reflected outside fo that function call.
This code illustrates the problem at https://play.golang.org/p/UebaqWNNJe or inlined
```go
package main

import "fmt"

type DB struct {
	m map[string]interface{}
}

func (db *DB) Set(key string, value interface{}) { db.m[key] = value }
func (db *DB) Get(key string) interface{}        { return db.m[key] }

func (db *DB) properClose() {
	for k := range db.m {
		delete(db.m, k)
	}
}

func (db *DB) buggyClose() {
	db = nil
}

func newDB() *DB { return &DB{m: make(map[string]interface{})} }

func main() {
	db1 := newDB()
	db1.Set("foo", "bar")
	fmt.Printf("Get: %v\n", db1.Get("foo"))
	db1.buggyClose()
	fmt.Printf("After buggy close: %v\n", db1.Get("foo"))
	db1.properClose()
	fmt.Printf("After proper close: %v\n", db1.Get("foo"))
}
```
invoking `db = nil` does nothing since that doesn't actually change the pointer data.

I found this while writing the tests for `tendermint/blockStore`

melekes:
Interesting find!

ebuchman:
Cool thanks!

PR #56, opened 2017-09-25
db: fix MemDB.Close

odeke-em:
Fixes https://github.com/tendermint/tmlibs/issues/55

MemDB previously mistakenly set the actual DB pointer to nil
although that side effect is not visible to the outside world
since p is an identifier within the scope of just that function
call. The proper fix is to delete the underlying data map.

### DIFF
--- a/db/mem_db.go
+++ b/db/mem_db.go
@@ -52,9 +52,11 @@ func (db *MemDB) DeleteSync(key []byte) {
 }
 
 func (db *MemDB) Close() {
-	db.mtx.Lock()
-	defer db.mtx.Unlock()
-	db = nil
+	// Close is a noop since for an in-memory
+	// database, we don't have a destination
+	// to flush contents to nor do we want
+	// any data loss on invoking Close()
+	// See the discussion in https://github.com/tendermint/tmlibs/pull/56
 }
 
 func (db *MemDB) Print() {
--- a/db/mem_db_test.go
+++ b/db/mem_db_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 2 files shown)


db/mem_db.go
```diff
@@ -53,8 +53,10 @@ func (db *MemDB) DeleteSync(key []byte) {
 
 func (db *MemDB) Close() {
 	db.mtx.Lock()
-	defer db.mtx.Unlock()
-	db = nil
+	for k := range db.db {
```
melekes: maybe `db = make(map[string][]byte)`?

melekes: I'd argue that we need to delete anything. If you close a "DB connection", your data aren't supposed to be deleted. So, for in-memory DB, the close should be empty IMHO.

odeke-em: So for other DBs, Close has different effects:
- CLevelDB: Close makes the DB unusable, so any other calls will panic https://github.com/jmhodges/levigo/blob/c42d9e0ca023e2198120196f842701bb4c55d7b9/db.go#L377-L388
- GoLevelDB: Close
https://github.com/tendermint/tmlibs/blob/9997e3a3b46db1d2f88aa9816ed0e7915dad6ac1/db/go_level_db.go#L81 --> Aborts all in-flight connections/requests and flushes all data. It advises that after close, the DB is unsafe to use


IMO an in-memory DB's close should be pretty much like flush, writing out all "unsynced" data. However, after considering your thoughts and a combination of the ideas from those DBs, I think perhaps we should instead just document that Close is a noop. If I want to actually delete all data, I can invoke db.Iterator and delete all fields. Thanks for the thoughts.

melekes: Are you gonna update this PR?

odeke-em: Yes, I am going to make it Close a noop

odeke-em:
All updated @melekes, PTAL.

PR #57, opened 2017-09-28
WIP: Improve db interface

cloudhead:
I would like to improve the DB interface to move some of the complexity down a layer, instead of having to deal with it when using the library. Currently, to achieve this, I have to do a type assertion and handle two cases manually (memdb and goleveldb). Also important to note that we don't expose Release() at the moment, which means the leveldb iterators could leak. We're going to be using the iterator feature a lot for historical queries, so this needs to be tight.

### DIFF
--- a/db/go_level_db.go
+++ b/db/go_level_db.go
@@ -121,6 +121,7 @@ type goLevelDBIterator struct {
 	iterator.Iterator
 }
 
+// Key returns a copy of the current key.
 func (it *goLevelDBIterator) Key() []byte {
 	key := it.Key()
 	k := make([]byte, len(key))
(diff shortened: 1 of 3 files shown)


db/go_level_db.go
```diff
@@ -115,8 +117,32 @@ func (db *GoLevelDB) Stats() map[string]string {
...
+}
+
+func (it *goLevelDBIterator) Value() []byte {
+	val := it.Value()
+	v := make([]byte, len(val))
+	copy(v, val)
+
+	return v
+}
+
```
ebuchman: Can we add comment explaining that the key/value will be copied

ebuchman:
This looks fine, but I'd note it's a breaking change

cloudhead:
@ebuchman in what way is it a breaking change? Do you mean because the interface has new methods? Are there other implementations of that interface?

ebuchman:
> because the interface has new methods

Right.

>  Are there other implementations of that interface

Not that I know of, but possibly something implemented by a user.

Regardless I think this can be merged. @melekes has been bumped to merge master for this repo, so I'll leave it to him :)

Issue #30, opened 2017-07-10
Add support for other dbs


ebuchman:
- https://github.com/dgraph-io/badger - lightning fast kv store in pure Go

odeke-em:
How about
- [ ] BoltDB

...Add others here

melekes:
There is an issue with too many dependencies. See

https://github.com/tendermint/tmlibs/pull/58#issuecomment-341873935
https://github.com/tendermint/tmlibs/pull/58#issuecomment-341901768
https://github.com/tendermint/tmlibs/pull/58#issuecomment-342107988

PR #58, opened 2017-09-30
STALLED: db: add BadgerDB

odeke-em:
Updates https://github.com/tendermint/tmlibs/issues/30

Adds a backend for BadgerDB as requested at
https://github.com/tendermint/tmlibs/issues/30#issue-241573231
and tests for it along.

BadgerDB has a few problems:
a) It doesn't have a BatchDelete, our fallback is thus Delete
because if do DeleteAsync, by the time batch.Write is invoked,
we can't be sure that the deletions have ran
b) Retrieving values from iterators requires a secondary allocation
equal to the length of the current key.
See https://godoc.org/github.com/dgraph-io/badger#KV.NewIterator
c) It seems to have arbitrary options for the size of the
log file varying between 10MB to 2GB. Currently with our
DB API, we can't pass in the required file size unless we
invoke:
```go
db.NewBadgerDBWithOptions(&db.Options{
  ValueLogFileSize: desiredFileSize,
})
```

Anyways, I've arbitrarily set the default value log file size to 1GB
given that Tendermint does a whole lot of heavy lifting with
keyvalue stores <-- cite my sources

### DIFF
--- a/db/badger_db.go
+++ b/db/badger_db.go
@@ -1,0 +1,323 @@
+// Copyright 2017 Tendermint. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// BadgerDB enables us the use of BadgerDB a fast key value store in pure Go.
+// Typical usage follows:
+//  db, err := db.NewBadgerDB("badger", targetDir)
+//  if err != nil {
+//	log.Fatalf("badgerDB initialization: %v", err)
+//  }
+//  defer db.Close()
+//
+//  For the asynchronous set
+//  db.Set([]byte("foo"), []byte("bar"))
+//  db.SetSync([]byte("true"), []byte("tendermint"))
+//  db.Delete([]byte("true"))
+//  db.Deletesync([]byte("true"))
+//  iter := db.Iterator()
+//  for {
+//     key, value := iter.Key(), iter.Value()
+//     fmt.Printf("key: %v value: %v\n", key, value)
+//  }
+//
+// For better performance, perform Sets/Writes in batches
+//  batch := db.NewBatch()
+//  for i := 0; i < 100; i++ {
+//	batch.Set(cmn.RandBytes(100), []byte("true!!!"))
+//  }
+//  batch.Write()
+
+package db
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"sync"
+
+	"github.com/dgraph-io/badger"
+
+	"github.com/tendermint/tmlibs/common"
+)
+
+func init() {
+	registerDBCreator(BadgerDBBackendStr, badgerDBCreator, true)
+}
+
+type Options badger.Options
+
+func badgerDBCreator(dbName, dir string) (DB, error) {
+	return NewBadgerDB(dbName, dir)
+}
+
+var (
+	_KB = int64(1024)
+	_MB = 1024 * _KB
+	_GB = 1024 * _MB
+)
+
+// NewBadgerDB creates a Badger key-value store backed to the
+// directory dir supplied. If dir does not exist, we create it.
+func NewBadgerDB(dbName, dir string) (*BadgerDB, error) {
+	// BadgerDB doesn't expose a way for us to
+	// create  a DB with the user's supplied name.
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return nil, err
+	}
+	opts := new(badger.Options)
+	*opts = badger.DefaultOptions
+	// Arbitrary size given that at Tendermint
+	// we'll need huge KeyValue stores.
+	opts.ValueLogFileSize = 1 * _GB
+	opts.Dir = dir
+	opts.ValueDir = dir
+
+	return NewBadgerDBWithOptions((*Options)(opts))
+}
+
+// NewBadgerDBWithOptions creates a BadgerDB key value store
+// gives the flexibility of initializing a database with the
+// respective options.
+func NewBadgerDBWithOptions(opts *Options) (*BadgerDB, error) {
+	kv, err := badger.NewKV((*badger.Options)(opts))
+	if err != nil {
+		return nil, err
+	}
+	return &BadgerDB{kv: kv}, nil
+}
+
+type BadgerDB struct {
+	kv *badger.KV
+}
+
+var _ DB = (*BadgerDB)(nil)
+
+func (b *BadgerDB) Get(key []byte) []byte {
+	valueItem := new(badger.KVItem)
+	if err := b.kv.Get(key, valueItem); err != nil {
+		// Unfortunate that Get can't return errors
+		// TODO: Propose allowing DB's Get to return errors too.
+		common.PanicCrisis(err)
+	}
+	var valueSave []byte
+	err := valueItem.Value(func(origValue []byte) error {
+		// TODO: Decide if we should just assign valueSave to origValue
+		// since here we aren't dealing with iterators directly.
+		valueSave = make([]byte, len(origValue))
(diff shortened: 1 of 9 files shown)


db/badger_db.go
```diff
@@ -0,0 +1,323 @@
...
+//  if err != nil {
+//	log.Fatalf("badgerDB initialization: %v", err)
+//  }
+//  defer db.Close()
+//
+//  For the asynchronous set
+//  db.Set([]byte("foo"), []byte("bar"))
+//  db.SetSync([]byte("true"), []byte("tendermint"))
+//  db.Delete([]byte("true"))
+//  db.Deletesync([]byte("true"))
```
melekes: `DeleteSync`

db/badger_db.go
```diff
@@ -0,0 +1,323 @@
...
+	_MB = 1024 * _KB
+	_GB = 1024 * _MB
+)
+
+// NewBadgerDB creates a Badger key-value store backed to the
+// directory dir supplied. If dir does not exist, we create it.
+func NewBadgerDB(dbName, dir string) (*BadgerDB, error) {
+	// BadgerDB doesn't expose a way for us to
+	// create  a DB with the user's supplied name.
+	if err := os.MkdirAll(dir, 0755); err != nil {
```
melekes: Don't you think `MkdirAll ` belongs to `NewBadgerDBWithOptions` function?

db/badger_db.go
```diff
@@ -0,0 +1,323 @@
...
+		return nil, err
+	}
+	opts := new(badger.Options)
```
melekes: what if we just do `opts := badger.DefaultOptions`?

db/badger_db.go
```diff
@@ -0,0 +1,323 @@
...
+type BadgerDB struct {
+	kv *badger.KV
+}
+
+var _ DB = (*BadgerDB)(nil)
+
+func (b *BadgerDB) Get(key []byte) []byte {
+	valueItem := new(badger.KVItem)
+	if err := b.kv.Get(key, valueItem); err != nil {
+		// Unfortunate that Get can't return errors
```
melekes: 💯

db/badger_db.go
```diff
@@ -0,0 +1,323 @@
...
+		// TODO: Propose allowing DB's Get to return errors too.
+		common.PanicCrisis(err)
+	}
+	var valueSave []byte
+	err := valueItem.Value(func(origValue []byte) error {
+		// TODO: Decide if we should just assign valueSave to origValue
```
melekes: so why can't we just assign?

db/badger_db.go
```diff
@@ -0,0 +1,323 @@
...
+	})
+	bb.entriesMu.Unlock()
+}
+
+// Unfortunately Badger doesn't have a batch delete
+// The closest that we can do is do a delete from the DB.
+// Hesitant to do DeleteAsync because that changes the
+// expected ordering
+func (bb *badgerDBBatch) Delete(key []byte) {
+	bb.db.Delete(key)
```
melekes: no error handling?

db/badger_db.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+// Write commits all batch sets to the DB
+func (bb *badgerDBBatch) Write() {
+	bb.entriesMu.Lock()
+	entries := bb.entries
+	bb.entries = nil
```
melekes: what if somebody calls `set` on this batch after this? code will panic

db/badger_db.go
odeke-em: Thanks, I'll update that.

db/badger_db.go
odeke-em: I don't think so, db.NewDB uses NewBadgerDB and expects the directory making to be handled magically. Moving it out of NewBadgerDB would mean that we'd have to document separately that BadgerDB needs to have its directory made before-hand, and that would make it awkward.

db/badger_db.go
odeke-em: For some reason, I initially thought `badger.DefaultOptions` was a pointer but somewhere I noticed not and left that in, lol. I'll update it, thank you!

ebuchman:
Cool! Can we base this on Alexis' branch (https://github.com/tendermint/tmlibs/pull/57) ? I'd love to see some benchmarks of performance vs our Go LevelDB

melekes:
also, tests are failing with a missing dependency error.

melekes:
> Cool! Can we base this on Alexis' branch (#57) ? I'd love to see some benchmarks of performance vs our Go LevelDB

alexis' branch was merged into develop. so this branch needs to be rebased against develop

melekes:
ok. I've rebased the PR. 

@odeke-em could you address some of my comments above and TODO here https://github.com/tendermint/tmlibs/pull/58/commits/2a54ceb94ba88370257622bea337b19cd3d5f378. 

Also, I would love to see benchmarks too https://github.com/tendermint/tmlibs/pull/58#issuecomment-333722132

melekes:
**I also have a concern that we're adding another dependency** (github.com/dgraph-io/badger) to our list of deps, which is already long. Is there a way we could only require it if it's being used?

melekes:
```
db/badger_db.go:277:22:warning: should use buf.String() instead of string(buf.Bytes()) (S1030) (megacheck)
db/badger_db.go:282:2:warning: field mu is unused (U1000) (megacheck)
db/badger_db.go:282:2:warning: unused struct field github.com/tendermint/tmlibs/db.badgerDBIterator.mu (structcheck)
db/badger_db_test.go:88::error: wrong number of args for format in Errorf call: 2 needed but 3 args (vet)
db/badger_db_test.go:186::error: wrong number of args for format in Errorf call: 2 needed but 3 args (vet)
db/example_test.go:24::error: Example_BadgerDB has malformed example suffix: BadgerDB (vet)
```

cloudhead:
If we're going to support a bunch of backends and don't want to pull in all the deps, we can do something like:

```
import "tmlibs/db/base"
import "tmlibs/db/drivers/badger"
import "tmlibs/db/drivers/leveldb"

func init() {
  base.Register("badger", badger.NewDriver())
  base.Register("leveldb", leveldb.NewDriver())
}
```

odeke-em:
@melekes I mean do we want the DB though? If we do, I don't see how we can avoid adding the dependency.

@cloudhead what do you mean by "not pull  in all the deps"? By including all those imports, the respective object code has to be compiled and linked. Also I believe we already have that style of registering backends https://github.com/tendermint/tmlibs/blob/d9525c0fb671204450b160807480e1263053fb20/db/db.go#L50-L64 but I perhaps don't understand what the concern here is, please help me understand it.

odeke-em:
@melekes thank you for the vets, I'll update them!

odeke-em:
@melekes fist bump for the vets and suggestions to make badgerDBIterator return an error, I've updated those points in commit https://github.com/tendermint/tmlibs/pull/58/commits/534cd7b90667687a289d0558676bc9b6545b7425

For the benchmarks, this cafe is as cold as Siberia, let me walk home, grab dinner and work on them there.

db/badger_db.go
odeke-em: Nope, after this, it starts a new batch. Take a look at the code in Set https://github.com/tendermint/tmlibs/pull/58/files#diff-9a61ac3e18917559e9062e23d4e9435dR237
or
screen shot 2017-11-05 at 2 01 10 pm

db/badger_db.go
odeke-em: The 'error handling' present with our structure doesn't return an error yet I am apprehensive about adding panics in Delete.

cloudhead:
@odeke-em sorry probably wasn't clear: the code snippet I wrote would happen user-side, not in tmlibs. So the user can import additional backends and register them with tmlibs/db. By default, tmlibs/db wouldn't come with any disk-based backends.

greg-szabo:
On today's (2018 feb 5th) tendermint dev meeting Bucky mentioned that according to Bucky's and Jae's research, BadgerDB is not mature enough for us to be used. Shall this PR be closed? If not, what's the priority / use case?

odeke-em:
Yeah @greg-szabo in deed, I raised some problems here https://github.com/tendermint/tmlibs/pull/58#issue-261880212 and also in the filed issue on BadgerDB https://github.com/dgraph-io/badger/issues/308, their API had drastically changed in the span of 2 weeks, but also that the max length of their keys is 1<<16 https://github.com/dgraph-io/badger/issues/308#issuecomment-342052570.

greg-szabo:
Closing this PR. Thanks for the work on it odeke, even though we pass on BadgerDB, it was an important point to raise and discuss it.

PR #62, opened 2017-10-16
fix goLevelDBIterator

melekes:
remove unnecessary return statement

### DIFF
--- a/db/go_level_db.go
+++ b/db/go_level_db.go
@@ -118,12 +118,12 @@ func (db *GoLevelDB) Stats() map[string]string {
 }
 
 type goLevelDBIterator struct {
-	iterator.Iterator
+	source iterator.Iterator
 }
 
 // Key returns a copy of the current key.
 func (it *goLevelDBIterator) Key() []byte {
-	key := it.Key()
+	key := it.source.Key()
 	k := make([]byte, len(key))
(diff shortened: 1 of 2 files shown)


melekes:
@odeke-em see https://github.com/tendermint/tmlibs/pull/62/commits/0b22b27bbb08947e3875dcfe22f95b983635eb0a . I updated the commit's description. Hope it's more clear right now.

db/go_level_db.go
```diff
@@ -118,12 +118,12 @@ func (db *GoLevelDB) Stats() map[string]string {
...
 
 type goLevelDBIterator struct {
-	iterator.Iterator
+	source iterator.Iterator
 }
 
 // Key returns a copy of the current key.
 func (it *goLevelDBIterator) Key() []byte {
-	key := it.Key()
+	key := it.source.Key()
```
ebuchman: oy can't believe we missed this first time around :(

ebuchman:
@odeke-em we had infinite loops before, so need a way to name the inner iterator. Could just refer to it as `.Iterator`, but an explicit name is great too. LGTM

db/go_level_db.go
cloudhead: I know right :/

PR #63, opened 2017-10-23
Make iterating over keys possible

wolfposd:


silasdavis:
Could make iterable:

```go
Iterate(consumer func(key string, value interface{}) (stop bool)) (stopped bool, err error)
```

melekes:
Care to write a small test? 

Btw, we should deprecate this in favor of https://godoc.org/sync#Map

common/cmap_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	cmap = NewCMap()
+)
+
+func TestIterateKeysWithValues(t *testing.T) {
+	cmap.Clear()
```
melekes: Self-contained tests are better in my view. Let's move `cmap = NewCMap()` inside each test

PR #64, opened 2017-10-23
Fix zeroed buffer getting flushed to the empty event

silasdavis:
You'll only notice this if you accidentally subscribe to the empty event (`""`) on Tendermint, but your `EventCache` implementation has a bug in that every time `Flush` is called it sends a prefix of 1000 empty events to the `EventSwitch` before sending the events it should have been buffering. Given this happens about every second it might not be an insignificant performance hit.

It looks like a guess was being made at a good size to have the buffer's backing array as 1000. I suggest we drop this as it is arbitrary and Go should do a good job of exponentially growing our arrays. If we do have a clear median on the number of events that will get buffered then it makes to set the backing array capacity to that number, but I suspect this is a premature optimisation. To do that we should use:

```go
evc.events = make([]eventInfo, 0, eventsBufferSize)
```

To give the backing array the desire capacity but set the length to 0 so we don't have a big empty prefix in our buffer and the first `append` adds at position 0.

### DIFF
--- a/events/event_cache.go
+++ b/events/event_cache.go
@@ -1,9 +1,5 @@
 package events
 
-const (
-	eventsBufferSize = 1000
-)
-
 // An EventCache buffers events for a Fireable
 // All events are cached. Filtering happens on Flush
 type EventCache struct {
@@ -14,8 +10,7 @@ type EventCache struct {
 // Create a new EventCache with an EventSwitch as backend
 func NewEventCache(evsw Fireable) *EventCache {
 	return &EventCache{
-		evsw:   evsw,
-		events: make([]eventInfo, eventsBufferSize),
+		evsw: evsw,
 	}
 }
 
@@ -27,7 +22,7 @@ type eventInfo struct {
 
 // Cache an event to be fired upon finality.
 func (evc *EventCache) FireEvent(event string, data EventData) {
-	// append to list
+	// append to list (go will grow our backing array exponentially)
 	evc.events = append(evc.events, eventInfo{event, data})
 }
 
@@ -37,5 +32,6 @@ func (evc *EventCache) Flush() {
 	for _, ei := range evc.events {
 		evc.evsw.FireEvent(ei.event, ei.data)
 	}
-	evc.events = make([]eventInfo, eventsBufferSize)
+	// Clear the buffer, since we only add to it with append it's safe to just set it to nil and maybe safe an allocation
+	evc.events = nil
 }
--- a/events/event_cache_test.go
+++ b/events/event_cache_test.go
@@ -1,0 +1,35 @@
+package events
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEventCache_Flush(t *testing.T) {
+	evsw := NewEventSwitch()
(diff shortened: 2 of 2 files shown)


melekes:
> You'll only notice this if you accidentally subscribe to the empty event ("") on Tendermint, but your EventCache implementation has a bug in that every time Flush is called it sends a prefix of 1000 empty events to the EventSwitch before sending the events it should have been buffering. Given this happens about every second it might not be an insignificant performance hit.

🤦‍♂️ omg.. thanks for finding this

Don't know why 1000 was chosen, but on average it should be equal to a number of transactions per sec, which is variable between 1 - 10000, so 1000 is probably a good number

silasdavis:
Okay see what you think of the approach in latest commit. We still start with an empty (nil) slice but instead of throwing it away we re-slice to zero length which hangs to whatever backing array has be allocated to hold the buffer, unless the current length is less than half the backing array capacity in which case it trims the capacity using the `[low:high:max]` syntax to the current length. This means it selects its own capacity adaptively.

We could just re-slice to 0 after each flush, which would hang on to all of the allocated capacity in the backing array. This might be fine but means it can only ever grow and if we ever have a particularly large spike in events we hang on to that memory permanently. I thought this would be a simple way of getting it to adapt without going totally over the top.

If the number of events between flushes are consistently within a factor of 2 then it will find the right capacity.

events/event_cache.go
```diff
@@ -37,5 +32,14 @@ func (evc *EventCache) Flush() {
 	for _, ei := range evc.events {
 		evc.evsw.FireEvent(ei.event, ei.data)
 	}
-	evc.events = make([]eventInfo, eventsBufferSize)
+	// Clear the buffer by re-slicing its length to zero
+	if cap(evc.events) > len(evc.events)<<1 {
```
ebuchman: is `x << 1` supposed to be better for some reason than `x * 2` ?

silasdavis: umm I think in my head it said 'order of magnitude', but on reflection its obfuscatory and certainly will compile down to the same thing

silasdavis:
I've squashed commits and added constant in place of the magic left shift

melekes:
Nooooooooo http://nooooooooooooooo.com/. I liked the first version more. The last one is too complex. And I am not sure it has to be that way because we haven't benchmarked this or anything.

silasdavis:
Which one do you mean when you say the first one?

melekes:
I cannot reference it anymore since you squshed the commits. Let's just do `evc.events = make([]eventInfo, 0, eventsBufferSize)`

silasdavis:
If you want something simepIe would suggest using:

```go
evc.events = evc.events[:0]
```
To keep the backing array, or:

```go
evc.events = nil
```
To throw it away.

Using a somewhat arbitrary made up buffer size is ugly.

melekes:
Can you restore the first commit?

silasdavis:
Yep there you go, that's the `nil` option, shall I drop the second commit?

melekes:
Please do

silasdavis:
done

silasdavis:
oh http://nooooooooooooooo.com/ is a real thing

melekes:
Awesome

PR #65, opened 2017-10-23
[autofile] Support for the new WAL format

melekes:
See https://github.com/tendermint/tendermint/pull/672
Issue: https://github.com/tendermint/tendermint/issues/573


melekes:
TODO:
- [ ] write tests

PR #67, opened 2017-10-24
CacheDB

jaekwon:
This PR is to make the tendermint/tmlibs/db CacheWrap() compatible.
We could in the future use cosmos/cosmos-sdk/state/kvcache.go (or whatever it becomes), but it's not clear whether we want to make the Cosmos KVStore/IterKVStore interfaces (designed for Merkle stores) compatible with the Tendermint DB interface (designed for LevelDB etc).  Until we can unify them lets keep them separate like this for now.

Also fixes a bug where MemDB keys weren't being ordered before Merkle-rooting.
Also fixes c-leveldb compatibility.  You need to run `brew install leveldb` or equivalent to run the tests.

NOTE: We *need* c-leveldb support so we should keep them as part of the tests, and fix our CI environment as necessary, instead of crippling our tests.


merkle/kvpairs.go
```diff
@@ -0,0 +1,49 @@
...
+type KVPairs []KVPair
+
+func (kvps KVPairs) Len() int           { return len(kvps) }
+func (kvps KVPairs) Less(i, j int) bool { return kvps[i].Key < kvps[j].Key }
+func (kvps KVPairs) Swap(i, j int)      { kvps[i], kvps[j] = kvps[j], kvps[i] }
+func (kvps KVPairs) Sort()              { sort.Sort(kvps) }
+
+func MakeSortedKVPairs(m map[string]interface{}) []Hashable {
+	kvPairs := []KVPair{}
+	for k, v := range m {
```
ebuchman: allocate up front

merkle/simple_map.go
```diff
@@ -0,0 +1,26 @@
...
+func (sm *SimpleMap) Set(k string, v interface{}) {
+	sm.kvz = append(sm.kvz, KVPair{Key: k, Value: v})
+}
+
+// Merkle root hash of items sorted by key.
+// NOTE: Behavior is undefined when key is duplicate.
+func (sm *SimpleMap) Hash() []byte {
+	sm.kvz.Sort()
+	kvPairsH := []Hashable{}
+	for _, kvp := range sm.kvz {
```
ebuchman: allocate up front

merkle/simple_map_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+func TestSimpleMap(t *testing.T) {
+	{
+		db := NewSimpleMap()
+		db.Set("key1", "value1")
+		assert.Equal(t, "376bf717ebe3659a34f68edb833dfdcf4a2d3c10", fmt.Sprintf("%x", db.Hash()), "Hash didn't match")
+	}
+	{
+		db := NewSimpleMap()
+		db.Set("key1", "value1")
```
ebuchman: duplicate test case ?

melekes:
`make metalinter_test` is failing with:

```
db/cache_db.go:134:33:warning: unnecessary conversion (unconvert)
db/backend_test.go:8:9:warning: unused variable or constant could not import github.com/tendermint/go-common (cannot find package "github.com/tendermint/go-common" in any of: (varcheck)
db/backend_test.go:17:18:warning: unused variable or constant undeclared name: common (varcheck)
db/cache_db.go:89::error: Println call ends with newline (vet)
db/cache_db.go:91::error: arg value for printf verb %X of wrong type: db.cDBValue (vet)
db/cache_db.go:93::error: Println call ends with newline (vet)
merkle/simple_proof.go:60::error: unreachable code (vet)
db/backend_test.go:8:9:warning: could not import github.com/tendermint/go-common (cannot find package "github.com/tendermint/go-common" in any of: (megacheck)
db/backend_test.go:17:18:warning: undeclared name: common (megacheck)
```

db/backend_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	common "github.com/tendermint/go-common"
+)
+
+func print(x []byte) {
```
melekes: don't see where it's being used...

db/c_level_db.go
```diff
@@ -150,3 +148,23 @@ func (mBatch *cLevelDBBatch) Write() {
...
+
+//----------------------------------------
+// CLevelDB Iterator adaptor
+
+type cLevelDBIterator struct {
+	*levigo.Iterator
+}
+
+func (c cLevelDBIterator) Next() bool {
+	c.Next()
```
melekes: infinite recursion? maybe `c.(*levigo.Iterator).Next()`

db/db.go
```diff
@@ -3,15 +3,14 @@ package db
 import . "github.com/tendermint/tmlibs/common"
 
 type DB interface {
-	Get([]byte) []byte
+	Get([]byte) []byte // NOTE: returns nil iff never set or deleted.
```
melekes: `if`

db/go_level_db.go
```diff
@@ -155,10 +154,6 @@ func (db *GoLevelDB) Iterator() Iterator {
 	return &goLevelDBIterator{db.db.NewIterator(nil, nil)}
 }
 
-func (db *GoLevelDB) IteratorPrefix(prefix []byte) Iterator {
```
melekes: why delete? nobody using it?

merkle/simple_map.go
melekes: and what if the value is not `Hashable`? I am asking because in `KVPair#Hash` there is an if `if kvH, ok := kv.Value.(Hashable); ok {`

adrianbrink:
`make metalinter_test` fails. Either we have to fix up all the warnings once or we disable it for now but run it locally and fix a little bit everytime.

db/db.go
```diff
@@ -3,15 +3,14 @@ package db
...
 	Set([]byte, []byte)
 	SetSync([]byte, []byte)
 	Delete([]byte)
 	DeleteSync([]byte)
 	Close()
 	NewBatch() Batch
 	Iterator() Iterator
-	IteratorPrefix([]byte) Iterator
```
cloudhead: We need this for iavl.

db/go_level_db.go
cloudhead: I'm using this in iavl.

merkle/simple_map_test.go
```diff
@@ -0,0 +1,48 @@
+package merkle
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	// "github.com/stretchr/testify/require"
```
cloudhead: We can delete this.

db/c_level_db.go
ebuchman: nice catch!

db/db.go
ebuchman: `iff` means `if and only if`

Makefile
```diff
@@ -33,15 +33,15 @@ metalinter_test: ensure_tools
 		--enable=goconst \
 		--enable=gosimple \
 	 	--enable=ineffassign \
-	   	--enable=interfacer \
+	  --enable=interfacer \
```
ebuchman: why did all this spacing get weird?

db/c_level_db.go
```diff
@@ -49,6 +51,8 @@ func NewCLevelDB(name string, dir string) (*CLevelDB, error) {
 	return database, nil
 }
 
+// Get gets the value for the given key. It returns ErrNotFound if the DB does not contains the
```
ebuchman: This doesn't sound round. The function panics on err, and should return nil if the key is not there. Is that not what it does?

db/c_level_db.go
```diff
@@ -19,13 +19,15 @@ func init() {
 	registerDBCreator(CLevelDBBackendStr, dbCreator, false)
 }
 
+// CLevelDB provides access to the C implementation of LevelDB
```
ebuchman: We should add a XXX comment that these methods panic on error from the underlying DB

db/db.go
melekes: did not know, thanks!

db/c_level_db.go
```diff
@@ -118,16 +133,39 @@ func (db *CLevelDB) Stats() map[string]string {
...
 
 //----------------------------------------
+// CLevelDB Iterator adaptor
+
+type cLevelDBIterator struct {
+	*levigo.Iterator
+}
+
+func (c cLevelDBIterator) Next() bool {
+	c.Next()
```
ebuchman: bad news bears

db/cache_db.go
```diff
@@ -37,63 +41,77 @@ func (db *CacheDB) Get(key []byte) []byte {
 	return db.parent.Get(key)
 }
 
+// Set sets the value for the given key in the cache. It does not write the value to the
+// underlying database. Set does not incur a performance penalty compared to SetNoLock.
```
ebuchman: What's all this about performance penalty? These functions lock a mutex while XxxNoLock does not

db/cache_db.go
```diff
@@ -37,63 +41,77 @@ func (db *CacheDB) Get(key []byte) []byte {
...
 
-// NOTE: Implements atomicSetDeleter
+// DeleteNoLock deletes the value for the given key. It does not delete the value from the
+// underlying database. DeleteNoLock implements atomicSetDeleter.
 func (db *CacheDB) DeleteNoLock(key []byte) {
 	db.cache[string(key)] = cDBValue{nil, true}
 }
 
+// Close closes the underlying database by deallocating the handle. Any attempts to use the DB
+// after close will result in a panic.
```
ebuchman: This probably isn't true, maybe we should clarify that the db should not be used after Close. But it won't panic in this case I don't thiknk ...

Makefile
adrianbrink: Mh, some had two spaces instead of tabs. Everything is now intended with 2 tabs.

db/c_level_db.go
adrianbrink: You are right. That doesn't sound right. I think I might have thought about something else while typing this :-)

db/cache_db.go
adrianbrink: That's right. It depends on the backing database.

db/go_level_db.go
jaekwon: I think we should take a look at levigo/goleveldb's iterators and what's supported there... first.

jaekwon: LeviGo supports Seek... https://github.com/jmhodges/levigo/blob/master/iterator.go#L121

jaekwon: So does Goleveldb: https://godoc.org/github.com/syndtr/goleveldb/leveldb/iterator#IteratorSeeker

Lets adopt these as tmlibs/db.Iterator, and make MemDB support it too...

db/c_level_db.go
jaekwon: c.Iterator.Next() should work...

db/go_level_db.go
jaekwon: I'll write an exposed module function called "IteratePrefix" that takes an iterator and calls Seek() accordingly to get the same effect.

.gitignore
```diff
@@ -1,4 +1,5 @@
 *.swp
```
odeke-em: We could also do
```shell
*.sw[op]
```
and it's awesome to know that you too @jaekwon use Vim!

Makefile
```diff
@@ -11,7 +11,7 @@ all: test
 NOVENDOR = go list github.com/tendermint/tmlibs/... | grep -v /vendor/
 
 test:
-	go test `glide novendor`
+	go test -tags gcc `glide novendor`
```
odeke-em: Am not too sure if we should always be including the gcc tag, since on my Mac this is going to trip out.

db/c_level_db.go
odeke-em: Or perhaps let's make it explicit to
```go
type cLevelDBIterator struct {
      iter *levigo.Iterator
}

func (c cLevelDBIterator) Next() bool {
      c.iter.Next()
}
```
to avoid such problems that we previously had with https://github.com/tendermint/tmlibs/commit/0b22b27bbb08947e3875dcfe22f95b983635eb0a where the problem slipped through the cracks?

db/cache_db.go
```diff
@@ -0,0 +1,172 @@
...
+// underlying database.
+func (db *CacheDB) Get(key []byte) []byte {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
+
+	value, ok := db.cache[string(key)]
+	if ok {
+		return value.value
+	}
+	return db.parent.Get(key)
```
odeke-em: Hmm, I don't see where we memoize the cache-missed value for this, perhaps let's do this.
```go
value, ok := db.cache[string(key)]
if !ok {
   data := db.parent.Get(key)
   value = cDBValue{value: data}
   db.cache[string(key)] = value
}
return value.value
```

db/cache_db.go
```diff
@@ -0,0 +1,172 @@
...
+// underlying database.
+func (db *CacheDB) SetSync(key []byte, value []byte) {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
+
+	db.SetNoLock(key, value)
+}
+
+// SetNoLock sets the value for the given key in the database. It does not write the value to the
+// underlying database. SetNoLock implements atomicSetDeleter.
```
odeke-em: How come we are referring to an `atomicSetDeleter` unexported interface?
Perhaps could we just do a compile type assertion and not mention it? e.g
```go
var _ atomicSetDeleter = (*CacheDB)(nil)
```

db/cache_db.go
odeke-em: Also should we perhaps check if `deleted { return nil }`?

db/cache_db.go
```diff
@@ -0,0 +1,172 @@
...
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
+
+	db.DeleteNoLock(key)
+}
+
+// DeleteNoLock deletes the value for the given key. It does not delete the value from the
+// underlying database. DeleteNoLock implements atomicSetDeleter.
+func (db *CacheDB) DeleteNoLock(key []byte) {
+	db.cache[string(key)] = cDBValue{nil, true}
```
odeke-em: I think we need mutex protection here :)

db/cache_db.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// Stats returns a map of property values for all keys and the size of the cache.
+func (db *CacheDB) Stats() map[string]string {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
+
+	stats := make(map[string]string)
+	stats["cache.size"] = fmt.Sprintf("%d", len(db.cache))
+	mergeStats(db.parent.Stats(), stats, "parent.")
```
odeke-em: Nice!

db/cache_db.go
```diff
@@ -0,0 +1,172 @@
...
+			batch.Delete([]byte(key))
+		} else {
+			batch.Set([]byte(key), []byte(value.value))
+		}
+	}
+	batch.Write()
+
+	// Clear the cache
+	db.cache = make(map[string]cDBValue)
+}
```
odeke-em: Very nice routine!

db/mem_db.go
```diff
@@ -12,46 +12,76 @@ func init() {
...
 	defer db.mtx.Unlock()
 	delete(db.db, string(key))
 }
 
+// DeleteNoLock deletes the value for the given key. It implements atomicSetDeleter.
+func (db *MemDB) DeleteNoLock(key []byte) {
+	delete(db.db, string(key))
+}
+
+// Close is a noop.
```
odeke-em: Thanks for the doc!

merkle/kvpairs.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	return hasher.Sum(nil)
+}
+
+type KVPairs []KVPair
+
+func (kvps KVPairs) Len() int           { return len(kvps) }
+func (kvps KVPairs) Less(i, j int) bool { return kvps[i].Key < kvps[j].Key }
+func (kvps KVPairs) Swap(i, j int)      { kvps[i], kvps[j] = kvps[j], kvps[i] }
+func (kvps KVPairs) Sort()              { sort.Sort(kvps) }
```
odeke-em: I just tagged you on the internal thread about the support for Go1.8+ where we'll later on just use `sort.Slice` that is
```go
sort.Slice(kvPairs, func(i, j int) bool {
   kvi, kvj := kvPairs[i], kvPairs[j]
   return kvi.Key < kvj.Key
})
```

merkle/simple_proof.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// Verify that leafHash is a leaf hash of the simple-merkle-tree
+// which hashes to rootHash.
+func (sp *SimpleProof) Verify(index int, total int, leafHash []byte, rootHash []byte) bool {
+	computedHash := computeHashFromAunts(index, total, leafHash, sp.Aunts)
+	if computedHash == nil {
+		return false
+	}
+	if !bytes.Equal(computedHash, rootHash) {
```
odeke-em:
```go
return computedHash != nil && bytes.Equal(computedHash, rootHash)
```

merkle/simple_proof.go
```diff
@@ -0,0 +1,138 @@
...
+// Use the leafHash and innerHashes to get the root merkle hash.
+// If the length of the innerHashes slice isn't exactly correct, the result is nil.
+func computeHashFromAunts(index int, total int, leafHash []byte, innerHashes [][]byte) []byte {
+	// Recursive impl.
+	if index >= total {
+		return nil
+	}
+	switch total {
+	case 0:
+		panic("Cannot call computeHashFromAunts() with 0 total")
```
odeke-em: And for a nerdy detour and if you get bored: this is one of my favorite compiler checks. We don't need the return nil below because `panic` is a terminating statement :) as per 
* https://golang.org/ref/spec#Return_statements
* https://golang.org/ref/spec#Terminating_statements
See https://play.golang.org/p/Uywqmi2m4G

Makefile
adrianbrink: How come it trips on your Mac? It runs fine on macOS 10.13 for me.

db/cache_db.go
adrianbrink: I was wondering about that too. Maybe @jaekwon can clarify how this functionality is supposed to work. I got the understanding that you had to explicitly add values to the cache.

db/cache_db.go
cloudhead: But the function is called `DeleteNoLock`

db/backend_test.go
ethanfrey: Didn't this repo die long ago? To be reborn as tmlibs

merkle/kvpairs.go
ethanfrey: iavl only compiles with go 1.8+
I questioned this, and bucky, alexis, and adrian all spoke in favor of making this a requirement. It is commented in cosmos-sdk (due to iavl dependency), let's just update everything to 1.8

db/cache_db.go
jaekwon: Doh, yeah.

db/cache_db.go
odeke-em: Ahah, thanks for that catch @cloudhead I made that suggestion without noticing that.

db/c_level_db.go
jaekwon: These comments are not necessary, we should just comment the interface functions (for DB and Iterator)

db/cache_db.go
jaekwon: if deleted, value.value will be nil.

db/cache_db.go
jaekwon: ok

merkle/simple_map.go
jaekwon: The KVPair is a Hashable that knows how to deal with binary objects.

Makefile
jaekwon: Works for me... go 1.8.1, 10.12.x

db/c_level_db.go
```diff
@@ -147,6 +139,65 @@ func (mBatch *cLevelDBBatch) Delete(key []byte) {
...
+
+func (c cLevelDBIterator) Prev() {
+	if !c.itr.Valid() {
+		panic("cLevelDBIterator Prev() called when invalid")
+	}
+	c.itr.Prev()
+}
+
+func (c cLevelDBIterator) Close() {
+	c.Close()
```
melekes: `c.itr.Close()`?

db/c_level_db.go
```diff
@@ -147,6 +139,65 @@ func (mBatch *cLevelDBBatch) Delete(key []byte) {
...
+}
+
+func (c cLevelDBIterator) GetError() error {
+	return c.GetError()
```
melekes: `c.itr.GetError()`?

db/c_level_db_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"fmt"
 	"testing"
 
-	. "github.com/tendermint/tmlibs/common"
+	"github.com/tendermint/tmlibs/common"
```
melekes: the convention is `cmn`

db/cache_db.go
```diff
@@ -0,0 +1,162 @@
...
+	return db
+}
+
+func (db *CacheDB) Get(key []byte) []byte {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
+
+	value, ok := db.cache[string(key)]
+	if !ok {
+		fmt.Println("1", string(key))
```
melekes: Are these debug statements?

db/common_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/tendermint/tmlibs/common"
+)
+
+func checkValid(t *testing.T, itr Iterator, expected bool) {
```
melekes: ❤️  these wrappers

db/common_test.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func checkNext(t *testing.T, itr Iterator, expected bool) {
+	itr.Next()
+	valid := itr.Valid()
+	assert.Equal(t, expected, valid)
+}
+
+func checkNextPanics(t *testing.T, itr Iterator) {
+	defer func() {
```
melekes: https://godoc.org/github.com/stretchr/testify/assert#Panics

```
assert.Panics(t, func(){ itr.Next() }, "Next expected panic but didn't")
```

db/common_test.go
```diff
@@ -0,0 +1,213 @@
...
+			// OK!
+		} else {
+			assert.Fail(t, "checkNextPanics expected panic but didn't")
+		}
+	}()
+	itr.Next()
+}
+
+func checkPrevPanics(t *testing.T, itr Iterator) {
+	defer func() {
```
melekes: same

db/common_test.go
```diff
@@ -0,0 +1,213 @@
...
+			// OK!
+		} else {
+			assert.Fail(t, "checkValuePanics expected panic but didn't")
+		}
+	}()
+	itr.Value()
+}
+
+func newTempDB(t *testing.T, backend string) (db DB) {
+	dir, dirname := common.Tempdir(fmt.Sprintf("test_go_iterator"))
```
melekes: no need for `fmt.Sprintf`

db/go_level_db.go
```diff
@@ -132,56 +172,50 @@ func (it *goLevelDBIterator) Key() []byte {
...
-type goLevelDBBatch struct {
-	db    *GoLevelDB
-	batch *leveldb.Batch
+func (it *goLevelDBIterator) Valid() bool {
+	if it.invalid {
+		return false
+	}
+	if !it.source.Valid() {
+		it.invalid = true
+		return false
```
melekes:
```
if it.invalid {
  return false
}
it.invalid = !it.source.Valid()
return !it.invalid
```

db/util.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (pi *prefixIterator) Value() []byte {
+	if pi.invalid {
+		panic("prefixIterator Value() called when invalid")
+	}
+	return pi.itr.Value()
+}
+
+func (pi *prefixIterator) Close()          {}
```
melekes: shouldn't we delegate to `pi.itr`? `pi.itr.Close()`

db/c_level_db.go
jaekwon: jesus.  thanks.

db/common_test.go
jaekwon: yay

Makefile
odeke-em: Aha nice, my Mac has issues with gcc tags but I'll try again.

db/c_level_db_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"fmt"
 	"testing"
 
-	. "github.com/tendermint/tmlibs/common"
+	cmn "github.com/tendermint/tmlibs/common"
```
odeke-em: Nice!

db/cache_db.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	return db
+}
+
+func (db *CacheDB) Get(key []byte) []byte {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
+
+	value, ok := db.cache[string(key)]
+	if !ok {
```
odeke-em: One other check we might need, I had mentioned in the previous review but in later comments: we perhaps want to check that value.deleted is not set lest that's a discarded value in the cache i.e
```go
if !ok || value.deleted {
}
```
What do y'all think?

db/cache_db.go
```diff
@@ -0,0 +1,157 @@
...
+		data := db.parent.Get(key)
+		value = cDBValue{value: data}
+		db.cache[string(key)] = value
+	}
+	return value.value
+}
+
+func (db *CacheDB) Set(key []byte, value []byte) {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
```
odeke-em: Now here I am being a bit painful about this, but I believe we can squeeze even better perhaps by avoiding defers whenever we can, as they are ~2X slower than non-defers, please see https://gist.github.com/odeke-em/f8fb3a8695b33dd06a700fc3473f3e62.
Could we please make all these:
```go
db.mtx.Lock()
db.SetNoLock(key, value)
db.mtx.Unlock()
```

db/cache_db.go
```diff
@@ -0,0 +1,157 @@
...
+
+	db.SetNoLock(key, value)
+}
+
+func (db *CacheDB) SetSync(key []byte, value []byte) {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
```
odeke-em: Ditto about the defers

db/cache_db.go
```diff
@@ -0,0 +1,157 @@
...
+	db.SetNoLock(key, value)
+}
+
+func (db *CacheDB) SetNoLock(key []byte, value []byte) {
+	db.cache[string(key)] = cDBValue{value, false}
+}
+
+func (db *CacheDB) Delete(key []byte) {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
```
odeke-em: Ditto about the defers

db/cache_db.go
```diff
@@ -0,0 +1,157 @@
...
+
+	db.DeleteNoLock(key)
+}
+
+func (db *CacheDB) DeleteSync(key []byte) {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
```
odeke-em: Ditto about the defers

db/cache_db.go
```diff
@@ -0,0 +1,157 @@
...
+	db.DeleteNoLock(key)
+}
+
+func (db *CacheDB) DeleteNoLock(key []byte) {
+	db.cache[string(key)] = cDBValue{nil, true}
+}
+
+func (db *CacheDB) Close() {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
```
odeke-em: Ditto about the defers

Makefile
odeke-em: This was where I first tripped out https://github.com/tendermint/merkleeyes/issues/39#issuecomment-322348864 2 months ago.

db/cache_db.go
```diff
@@ -0,0 +1,157 @@
...
+
+	db.parent.Close()
+}
+
+func (db *CacheDB) Print() {
```
odeke-em: So while Print is useful from an external perspective, it is not that easy to test. IMHO we could do something like
```go
func (db *CacheDB) Fprint(w io.Writer) {
   db.mtx.Lock()
   defer db.mtx.Unlock()

   fmt.Fprintln(w, "CacheDB\ncache:")
   for key, value := range db.cache {
         fmt.Fprintf(w, "[%X]:\t[%v\n", key, value)
   }
   fmt.Fprintln(w, "\nparent:")
   db.parent.Print()
}

func (db *CacheDB) Print() { db.Fprint(os.Stdout) }
```

and then properly test the order and expected out.

db/db.go
```diff
@@ -24,17 +25,59 @@ type Batch interface {
 	Write()
 }
 
+/*
+	Usage:
```
odeke-em: Me likey!

db/go_level_db_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"fmt"
 	"testing"
 
-	. "github.com/tendermint/tmlibs/common"
+	"github.com/tendermint/tmlibs/common"
```
odeke-em: Minor nit, the convention seemingly used is `cmn` instead of `common` ie
```go
cmn "github.com/tendermint/tmlibs/common"
```

db/mem_batch.go
```diff
@@ -0,0 +1,50 @@
...
+
+type memBatch struct {
+	db  atomicSetDeleter
+	ops []operation
+}
+
+type opType int
+
+const (
+	opTypeSet    = 1
```
odeke-em: We missed setting the opType here so
```go
const (
   opTypeSet opType = 1
   opTypeDelete opType = 2
)
```

db/mem_batch.go
```diff
@@ -0,0 +1,50 @@
...
+func (mBatch *memBatch) Set(key, value []byte) {
+	mBatch.ops = append(mBatch.ops, operation{opTypeSet, key, value})
+}
+
+func (mBatch *memBatch) Delete(key []byte) {
+	mBatch.ops = append(mBatch.ops, operation{opTypeDelete, key, nil})
+}
+
+func (mBatch *memBatch) Write() {
+	mtx := mBatch.db.Mutex()
```
odeke-em: Nice!

db/mem_batch.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (mBatch *memBatch) Set(key, value []byte) {
+	mBatch.ops = append(mBatch.ops, operation{opTypeSet, key, value})
+}
+
+func (mBatch *memBatch) Delete(key []byte) {
+	mBatch.ops = append(mBatch.ops, operation{opTypeDelete, key, nil})
+}
+
+func (mBatch *memBatch) Write() {
```
odeke-em: Write? Flush perhaps? Am just bikeshedding on the name, it doesn't matter much :)

db/mem_db.go
```diff
@@ -28,15 +29,30 @@ func (db *MemDB) Get(key []byte) []byte {
 	return db.db[string(key)]
 }
 
+func (db *MemDB) GetOK(key []byte) ([]byte, bool) {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
```
odeke-em: Ditto here about an unnecessary defer.

db/mem_db.go
```diff
@@ -28,15 +29,30 @@ func (db *MemDB) Get(key []byte) []byte {
...
 	db.mtx.Lock()
 	defer db.mtx.Unlock()
-	db.db[string(key)] = value
+	db.SetNoLock(key, value)
 }
 
 func (db *MemDB) SetSync(key []byte, value []byte) {
 	db.mtx.Lock()
 	defer db.mtx.Unlock()
+	db.SetNoLock(key, value)
```
odeke-em: Ditto here about the unnecessary defer.

db/db.go
```diff
@@ -24,17 +28,59 @@ type Batch interface {
...
+
+	for itr.Seek(mykey); itr.Valid(); itr.Next() {
+		k, v := itr.Key(); itr.Value()
+		....
+	}
+*/
 type Iterator interface {
-	Next() bool
 
+	// Seek moves the iterator the position of the key given or, if the key
```
ethanfrey: So seek is like the replacement for IteratorPrefix?
Can you add in the example a check to show how to iterate in a range? Cuz if we use seek to find the beginning, there is probably an end we also want to see.

But yeah, Seek() with adding Prev() as well as  Next() seems like a nice generalization that allows a lot more flexibility to whoever uses this. Just show those example usages.

db/db.go
```diff
@@ -3,19 +3,23 @@ package db
...
 	Iterator() Iterator
-	IteratorPrefix([]byte) Iterator
 
 	// For debugging
 	Print()
+
+	// Stats returns a map of property values for all keys and the size of the cache.
 	Stats() map[string]string
+
+	// CacheWrap wraps the DB w/ a CacheDB.
```
ethanfrey: Why does this return interface{}? It provides no guarantees.

In cachedb_test.go, I see this line often:
`cdb := mem.CacheWrap().(*CacheDB)`

If the only way to use it is to cast it to a particular type, then just force that type.
If you want people to be able to use the CacheWrap without forcing a given type, then return a usable interface,that exposes some subset of the DB interface.

ethanfrey: Actually, it seems *CacheDB implements the entire DB interface, why not return it there? Or remove a method or two that you don't want to require from Caches...

```Go
type Cache interface {
.....
}

type DB interface {
  Cache
  NewBatch() Batch
  Iterate() Iterator
}
```

then we could just use `cdb := mem.CacheWrap()`

db/cache_db_test.go
```diff
@@ -0,0 +1,83 @@
...
+	cdb := mem.CacheWrap().(*CacheDB)
+	cdb2 := cdb.CacheWrap().(*CacheDB)
+	require.NotPanics(t, func() { cdb2.Write() })
+	require.Panics(t, func() { cdb2.Write() })
+	cdb2 = cdb.CacheWrap().(*CacheDB)
+	require.NotPanics(t, func() { cdb2.Write() })
+	require.Panics(t, func() { cdb2.Write() })
+}
+
+func TestCacheDBNested(t *testing.T) {
```
ethanfrey: Good tests. Maybe add a few more.

What if we make two caches A, and B from one parent, and then try to write them both?
Is this allowed? Does it return an error? If it is allowed, what is supposed to happen?

Test cases are a great way to document and enforce this case.

db/cache_db.go
```diff
@@ -0,0 +1,227 @@
...
+// If value is nil but deleted is false,
+// it means the parent doesn't have the key.
+// (No need to delete upon Write())
+type cDBValue struct {
+	value   []byte
+	deleted bool
+}
+
+// CacheDB wraps an in-memory cache around an underlying DB.
+type CacheDB struct {
```
ethanfrey: We really should have two caches, a read cache and a write cache... Otherwise, if we query a value, we write that to the db again. Not only is this wasteful use of IO, but with the version tree, this actually changes the hash.

https://github.com/cosmos/cosmos-sdk/blob/develop/state/kvcache.go#L6-L11

Issue:
https://github.com/cosmos/gaia/issues/41
https://github.com/cosmos/cosmos-sdk/pull/271

db/cache_db.go
jaekwon: I insist that we keep the defers until we find that it's a bottleneck... it's defensive programming that IMO is well worth it.

db/cache_db.go
jaekwon: not value.deleted... if value.deleted in the cache, we don't want to query the base store.

if value.deleted, value.value is nil btw.

db/cache_db.go
jaekwon: I wouldn't want to test Print functions TBH.

db/mem_batch.go
jaekwon: You mean you're yak-shaving the lingo?

db/cache_db.go
jaekwon: Fixing so that we have a "dirty" flag.

db/db.go
jaekwon: Maybe.  But CacheWrap also exists in the SDK for the merkle stores, and those don't conform to tmlibs/db.DB...

db/db.go
jaekwon: No, the replacement for IteratorPrefix is db.IteratePrefix(db DB, prefix []byte) Iterator.

jaekwon: Besides using db.IteratePrefix, You'll just have to manually stop iteration when you reach the end that you desire

Issue #69, opened 2017-10-26
autofile: groupReader.Read tests and impl


ebuchman:
https://github.com/tendermint/tendermint/pull/672#discussion_r147146477

PR #71, opened 2017-10-30
[autofile] test GroupReader more extensively (Refs #69)

melekes:


autofile/group.go
```diff
@@ -596,14 +596,12 @@ func (gr *GroupReader) Read(p []byte) (n int, err error) {
...
 		if err == io.EOF {
-			// Open the next file
-			if err1 := gr.openFile(gr.curIndex + 1); err1 != nil {
-				return n, err1
-			}
 			if n >= lenP {
 				return n, nil
-			} else {
-				continue
+			} else { // Open the next file
```
ebuchman: does this actually make a difference re correctness? or we're just saving ourselves from unnecessarily opening the next file sometimes?

melekes: yes, this is more correct. opening of next file may result in an error (?). plus "saving ourselves from unnecessarily opening the next file sometimes", of course

PR #72, opened 2017-11-01
Fsdb

jaekwon:
Uses the filesystem, keys are just URL encoded keys in the directory.
This will be used to store encrypted user keys, from go-crypto/keys.


db/fsdb.go
```diff
@@ -0,0 +1,223 @@
...
+		panic(errors.Wrap(err, fmt.Sprintf("Listing keys in %s", db.dir)))
+	}
+	sort.Strings(keys)
+	it.keys = keys
+	return it
+}
+
+func (db *FSDB) nameToPath(name []byte) string {
+	n := url.PathEscape(string(name))
+	return path.Join(db.dir, n)
```
melekes: `filepath.Join`. `path.Join` does not work for Windows

db/fsdb.go
```diff
@@ -0,0 +1,223 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer dir.Close()
+
+	names, err := dir.Readdirnames(0)
+	for i, name := range names {
+		n, err := url.PathUnescape(name)
+		if err != nil {
+			return nil, errors.New(fmt.Sprintf("Failed to unescape %s while listing", name))
```
melekes: `fmt.Errorf("Failed to unescape %s while listing", name)`

db/fsdb.go
```diff
@@ -0,0 +1,223 @@
...
+
+// Write some bytes from a file
+// CONTRACT: returns os errors directly without wrapping
+func write(path string, d []byte) error {
+	f, err := os.OpenFile(path, os.O_CREATE|os.O_EXCL|os.O_WRONLY, keyPerm)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	_, err = f.Write(d)
```
melekes: Should we call `f.Sync()` after?

db/fsdb.go
```diff
@@ -0,0 +1,223 @@
...
+	defer f.Close()
+
+	d, err := ioutil.ReadAll(f)
+	if err != nil {
+		return nil, err
+	}
+	return d, nil
+}
+
+// Write some bytes from a file
```
melekes: `to a file`

db/db.go
```diff
@@ -87,6 +87,7 @@ const (
 	CLevelDBBackendStr  = "cleveldb"
 	GoLevelDBBackendStr = "goleveldb"
 	MemDBBackendStr     = "memdb"
+	FSDBBackendStr      = "fsdb" // using the filesystem naively
```
melekes: why do need it? just curios

jaekwon: We had similar code in go-crypto/keys/* to save files in the filesystem (or a mock memory system for testing).  With this we can use the DB interface, instead of everyone writing their own unit tests to test io stuff for their custom X-manager  e.g. key-manager).

db/fsdb.go
jaekwon: awesome

db/fsdb.go
jaekwon: yeah great catch.

db/db.go
melekes: Ok, cool. Thanks for explaining this to me.

Issue #45, opened 2017-08-16
common.Service: review Start's return signature (bool, error)


ebuchman:
https://github.com/tendermint/go-common/issues/1

ebuchman:
Why bother returning a bool? Is the error not enough?

https://github.com/tendermint/go-common/blob/master/service.go#L78
 @melekes
     Member
melekes commented on Mar 21
yeah, bool is superfluous
 @ethanfrey
     Member
ethanfrey commented on Mar 21
If I remember correctly when I was writing test code, if I call Start() on a Service that is already running, it returns (false, nil). Only if I try to legitimately start it, but it fails in startup do I get an error.

The distinction is quite important to make it safe for reentrant calls. The other approach would be to have a special error type like ErrAlreadyStarted, then check for that in your code explicitly. Kind of like if I make a db call in gorm, and get an error, I check if it is a RecordNotFound error, or whether there was a real error with the db query.
 @melekes
     Member
melekes commented on Mar 21
Ah, I see. Thanks. I must say I like ErrAlreadyStarted approach more (not just in Golang)

odeke-em:
I've retitled the issue a little.

IMHO in deed, like @melekes says, returning bool is unnecessary. I'd support the approach of just returning a concrete error like @ethanfrey suggested with `ErrAlreadyStarted` thereby replacing `(bool, error)` with just `error`

### Usage
```go
switch err := svc.Start(); err != nil {
  case common.ErrAlreadyStarted:
  case common.ErrAlreadyStopped:
  default:
    // Miscellaneous error handle it
    return nil, err
  case nil:
    // Home free
}
```

Also I think we can make the code clearer with a

PR #75, opened 2017-11-06
change common start signature

melekes:
Refs #45


ebuchman:
I feel like Jae had a reason for wanting the bool here. We should check in with him.

melekes:
> And should we add an error to OnStop?

I don't think we want to support that. like if there is an error in OnStop code, it should be ignored or dealt immediately? i.e. this should not bubble up

> Should we also do this for Reset ?

ye

common/service.go
```diff
@@ -139,17 +146,17 @@ func (bs *BaseService) Stop() bool {
 func (bs *BaseService) OnStop() {}
 
 // Implements Service
-func (bs *BaseService) Reset() (bool, error) {
+func (bs *BaseService) Reset() error {
 	if !atomic.CompareAndSwapUint32(&bs.stopped, 1, 0) {
 		bs.Logger.Debug(Fmt("Can't reset %v. Not stopped", bs.name), "impl", bs.impl)
-		return false, nil
+		return fmt.Errorf("can't reset running %s", bs.name)
```
melekes: maybe we should `panic` here. like clearly if we are trying to reset running service, then something wrong with our code?

ebuchman:
> I don't think we want to support that. like if there is an error in OnStop code, it should be ignored or dealt immediately? i.e. this should not bubble up

But we just changed `Stop` to return an `error`. Wouldn't it be misleading if it returned `nil` but some piece of the service didn't stop because `OnStop` had an error ? 

I agree it's not something that should happen, and OnStop should deal as best it can, but just trying to be consistent and unsurprising.

common/service.go
ebuchman: We could, but it's probably best if we try to minimize panics in libraries ?

adrianbrink:
I think it should bubble up from OnStop. A user should be able to catch an error and return it in his implementation of `OnStop`.

melekes:
> but some piece of the service didn't stop because OnStop had an error ?

OnStop should never result in an error. One example could be Erlang gen_server http://erlang.org/doc/design_principles/gen_server_concepts.html#id64968 (notice that terminate always returns `ok`)

> but just trying to be consistent and unsurprising.

what if we add a comment about that?

ebuchman:
> OnStop should never result in an error

What if it's supposed to do some cleanup and the cleanup fails ? I guess this means it's a completely contained responsibility. I suppose that's ok.

Makefile
```diff
@@ -4,14 +4,18 @@ GOTOOLS = \
 	github.com/Masterminds/glide \
 	github.com/alecthomas/gometalinter
 
+PACKAGES=$(shell go list ./... | grep -v '/vendor/')
 REPO:=github.com/tendermint/tmlibs
 
 all: test
 
 NOVENDOR = go list github.com/tendermint/tmlibs/... | grep -v /vendor/
```
melekes: this line can be removed now (it is identical to `PACKAGES`)

PR #82, opened 2017-11-27
Remove SetLogger from Service interface

adrianbrink:
This is a breaking change and will require that all downstream clients upgrade. This is possibly one of the last times where we can make this kind of change to the base apis. 

The main reason is:
```
Unless you know which concrete object (which implements service) you are dealing with you cannot set a sensible logger.

There is never a point, where you iterate through a bunch of Service objects and call `SetLogger` on all of them. 
```

From slack:
```
adrian [10:56] 
@bucky @melekes Is there a particular reason why we always set the loggers using `SetLogger` instead of passing them in the constructor?

ethan.frey [11:01] 
AFAIK it was historical to not break the api by adding another object to the constructors for some experimental logging overhaul.
IMO it is time to make logger a first class citizen
It was for an easy transition from the old format

adrian [11:02] 
okay. Is there a use case to make the logger settable after instantiation? So if it's passed into the constructor do we still want to maintain the `SetLogger` methods?

adrian [13:00] 
For review: https://github.com/tendermint/tendermint/pull/898

adrian [13:24] 
In tmlibs the Service interface has SetLogger. What is the purpose of it? Unless you know which concrete object (which implements service) you are dealing with you cannot set a sensible logger. I'd be in favour of removing SetLogger from the Service interface.
 
melekes [16:23] 
> Unless you know which concrete object (which implements service) you are dealing with you cannot set a sensible logger

that's why all our services have logger and you only set it if you need logging for that service

> Is there a particular reason why we always set the loggers using `SetLogger` instead of passing them in the constructor?
I don't like when logger is one of the params. you can disagree, but I think constructor params should be only for absolutely necessary things and logger is not necessary (like service can be perfectly fine without a logger) (edited)

adrian [16:35] 
Mh, I'd be okay to also remove it from the constructors. But it feels like `SetLogger` should not be part of the service interface. There is never a point, where you iterate through a bunch of Service objects and call `SetLogger` on all of them. 
In essence you only call `SetLogger` on concrete structs.

melekes [17:02] 
> But it feels like `SetLogger` should not be part of the service interface
I kinda feel the same. Maybe we can still remove it
```

### DIFF
--- a/common/service.go
+++ b/common/service.go
@@ -6,6 +6,7 @@ import (
 	"github.com/tendermint/tmlibs/log"
 )
 
+// Service describes something that can be started, stopped and reset.
 type Service interface {
 	Start() (bool, error)
 	OnStart() error
@@ -19,8 +20,6 @@ type Service interface {
 	IsRunning() bool
 
 	String() string
-
-	SetLogger(log.Logger)
 }
 
 /*
@@ -65,18 +64,23 @@ Typical usage:
 		// stop subroutines, etc.
 	}
 */
+
+// BaseService implements Service.
 type BaseService struct {
-	Logger  log.Logger
+	Logger log.Logger
+	Quit   chan struct{}
+
 	name    string
 	started uint32 // atomic
 	stopped uint32 // atomic
-	Quit    chan struct{}
 
 	// The "subclass" of BaseService
 	impl Service
 }
 
-func NewBaseService(logger log.Logger, name string, impl Service) *BaseService {
+// NewBaseService returns a pointer to a newly instantiated BaseService object.
+// logger can be nil and will then default to a NewNopLogger.
+func NewBaseService(name string, impl Service, logger log.Logger) *BaseService {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	}
@@ -89,11 +93,7 @@ func NewBaseService(logger log.Logger, name string, impl Service) *BaseService {
 	}
 }
 
-func (bs *BaseService) SetLogger(l log.Logger) {
-	bs.Logger = l
-}
-
-// Implements Servce
+// Start implements Service
 func (bs *BaseService) Start() (bool, error) {
 	if atomic.CompareAndSwapUint32(&bs.started, 0, 1) {
 		if atomic.LoadUint32(&bs.stopped) == 1 {
@@ -115,12 +115,12 @@ func (bs *BaseService) Start() (bool, error) {
 	}
 }
 
-// Implements Service
+// OnStart implements Service
 // NOTE: Do not put anything in here,
 // that way users don't need to call BaseService.OnStart()
 func (bs *BaseService) OnStart() error { return nil }
 
-// Implements Service
+// Stop implements Service
 func (bs *BaseService) Stop() bool {
 	if atomic.CompareAndSwapUint32(&bs.stopped, 0, 1) {
 		bs.Logger.Info(Fmt("Stopping %v", bs.name), "impl", bs.impl)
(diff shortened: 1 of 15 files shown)


common/service.go
```diff
@@ -65,18 +64,23 @@ Typical usage:
...
-	Quit    chan struct{}
 
 	// The "subclass" of BaseService
 	impl Service
 }
 
-func NewBaseService(logger log.Logger, name string, impl Service) *BaseService {
+// NewBaseService returns a pointer to a newly instantiated BaseService object.
+// logger can be nil and will then default to a NewNopLogger.
+func NewBaseService(name string, impl Service, logger log.Logger) *BaseService {
```
ebuchman: unless there's really good reason we shouldn't go changing the order of parameters ... its just a big nuisance downstream

ebuchman:
Can you take a look through tendermint and see how prevalent the use of SetLogger is and if there's anywhere it's actually helpful? If you can help update Tendermint for this change, we can consider moving forward with it - but it will be a big one

common/service.go
adrianbrink: Yeah, I was thinking about this one too.

Actually I came across a situation where it makes sense to keep it in front. I'll change it to the front.

I just reordered it for style :-)

adrianbrink:
It is very prevalent in Core right now, but it is almost exclusively used directly after the type constructor. 

I'll update it in Tendermint core too. This is the first step: https://github.com/tendermint/tendermint/pull/898

adrianbrink:
We don't even have to remove all instances of `SetLogger`. We just remove it from the interface, which means that it is only callable on concrete types instead of on the interface types.

pubsub/example_test.go
```diff
@@ -12,8 +12,7 @@ import (
 )
 
 func TestExample(t *testing.T) {
-	s := pubsub.NewServer()
-	s.SetLogger(log.TestingLogger())
+	s := pubsub.NewServer(log.TestingLogger())
```
melekes: I prefer `SetLogger` approach

pubsub/pubsub_test.go
```diff
@@ -184,18 +183,21 @@ func benchmarkNClients(n int, b *testing.B) {
...
 	for i := 0; i < b.N; i++ {
-		s.PublishWithTags(ctx, "Gamora", map[string]interface{}{"abci.Account.Owner": "Ivan", "abci.Invoices.Number": i})
+		s.PublishWithTags(ctx, "Gamora",
+			map[string]interface{}{"abci.Account.Owner": "Ivan", "abci.Invoices.Number": i})
 	}
 }
 
 func benchmarkNClientsOneQuery(n int, b *testing.B) {
-	s := pubsub.NewServer()
+	s := pubsub.NewServer(nil)
```
melekes: and I don't like this... when some dev looks at this line, how should he guess what nil is??? solution could be using options (like `pubsub.NewServer()` & `pubsub.NewServer(Logger(log.TestingLogger()))`). But I don't see how current `SetLogger()` is different.

melekes:
> Unless you know which concrete object (which implements service) you are dealing with you cannot set a sensible logger.

completely agree. but as for the rest, I am against moving logger into a constructor.

pubsub/pubsub_test.go
melekes: or we can pass `log.NopLogger()`, but not nil, please

ethanfrey:
How about SetLogger returning a Service as well, so we can chain it in the constructors. Then everyone should be happy.

S := NewPubSub().SetLogger(logger)

PR #84, opened 2017-11-29
query#Conditions for indexing

melekes:


melekes:
```
WARNING: staticcheck, gosimple and unused are all set, using megacheck instead
clist/clist.go:96::warning: Use of unsafe calls should be audited,LOW,HIGH (gas)
clist/clist.go:112::warning: Use of unsafe calls should be audited,LOW,HIGH (gas)
common/os.go:27::warning: Subprocess launching with partial path.,MEDIUM,HIGH (gas)
common/random.go:60::warning: Use of weak random number generator (math/rand instead of crypto/rand),HIGH,MEDIUM (gas)
common/random.go:76::warning: Use of weak random number generator (math/rand instead of crypto/rand),HIGH,MEDIUM (gas)
process/process.go:28::warning: Subprocess launching with variable.,HIGH,HIGH (gas)
make: *** [metalinter_test] Error 1
```

ebuchman:
Just merged the other PR - lets rebase this on develop. Turned off that linter for now, we can wait for @zramsay to come around and fix.

pubsub/query/query.go
```diff
@@ -48,17 +56,91 @@ func (q *Query) String() string {
 	return q.str
 }
 
-type operator uint8
+// Operator is an operator that defines some kind of relation between tag and
+// operand (equality, etc.).
+type Operator uint8
```
melekes: do we want to expose that? what if we return operator as a string?

PR #94, opened 2017-12-06
Refactor throttle timer

ethanfrey:
As I was reviewing abci, I came across throttle timer, which seemed very confusing code to me, in that it mixed mutex and channel, and accessed state across multiple goroutines (some generated internally to the time package via AfterFunc). I added tests and proved to myself it worked properly, but I wanted to try my hand at refactoring it. One for cleaner code, and also for me to get more comfortable with how goroutines and channels are used in tendermint code.

I would like a review with one of three outcomes:

* The new code is broken - here's why
* The new code works, but it is not better than the old code - here's why
* The new code works, and is cleaner than the old code

In any case, it will be a learning experience for me. If you like it, merge it. If not, please explain why the previous state was better, so I can write similar code.

### DIFF
--- a/common/throttle_timer.go
+++ b/common/throttle_timer.go
@@ -30,7 +30,7 @@ const (
 )
 
 func NewThrottleTimer(name string, dur time.Duration) *ThrottleTimer {
-	c := make(chan struct{}, 1)
+	c := make(chan struct{})
 	var t = &ThrottleTimer{
 		Name:   name,
 		Ch:     c,
@@ -54,9 +54,19 @@ func (t *ThrottleTimer) run() {
 				return
 			}
 		case <-t.timer.C:
-			t.isSet = false
-			t.output <- struct{}{}
+			t.trySend()
 		}
+	}
+}
+
+// trySend performs non-blocking send on t.output (t.Ch)
+func (t *ThrottleTimer) trySend() {
+	select {
+	case t.output <- struct{}{}:
+		t.isSet = false
+	default:
+		// if we just want to drop, replace this with t.isSet = false
+		t.timer.Reset(t.dur)
 	}
 }
 
--- a/common/throttle_timer.go
+++ b/common/throttle_timer.go
@@ -1,7 +1,6 @@
 package common
 
 import (
-	"fmt"
 	"time"
 )
 
@@ -64,7 +63,6 @@ func (t *ThrottleTimer) run() {
 // happen in this method. It is only called from the run goroutine
 // so we avoid any race conditions
 func (t *ThrottleTimer) processInput(cmd command) (shutdown bool) {
-	fmt.Printf("processInput: %d\n", cmd)
 	switch cmd {
 	case Set:
 		if !t.isSet {
@@ -77,9 +75,7 @@ func (t *ThrottleTimer) processInput(cmd command) (shutdown bool) {
 	case Unset:
 		if t.isSet {
 			t.isSet = false
-			if !t.timer.Stop() {
-				<-t.timer.C
-			}
+			t.timer.Stop()
 		}
 	default:
 		panic("unknown command!")
--- a/common/throttle_timer.go
+++ b/common/throttle_timer.go
@@ -52,8 +52,8 @@ func (t *ThrottleTimer) run() {
 		select {
 		case cmd := <-t.input:
 			// stop goroutine if the input says so
+			// don't close channels, as closed channels mess up select reads
 			if t.processInput(cmd) {
-				close(t.Ch)
 				return
 			}
 		case <-t.timer.C:
--- a/common/throttle_timer.go
+++ b/common/throttle_timer.go
@@ -11,11 +11,10 @@ If a long continuous burst of .Set() calls happens, ThrottleTimer fires
 at most once every "dur".
 */
 type ThrottleTimer struct {
-	Name   string
-	Ch     <-chan struct{}
-	output chan<- struct{}
-	input  chan command
-	dur    time.Duration
+	Name  string
+	Ch    chan struct{}
+	input chan command
+	dur   time.Duration
 
 	timer *time.Timer
 	isSet bool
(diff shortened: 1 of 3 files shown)


common/throttle_timer.go
```diff
@@ -12,54 +11,85 @@ If a long continuous burst of .Set() calls happens, ThrottleTimer fires
 at most once every "dur".
 */
 type ThrottleTimer struct {
-	Name string
-	Ch   chan struct{}
```
ebuchman: lets make it unidirectional so it's clear

common/throttle_timer.go
```diff
@@ -12,54 +11,85 @@ If a long continuous burst of .Set() calls happens, ThrottleTimer fires
...
+			// stop goroutine if the input says so
+			if t.processInput(cmd) {
+				// TODO: do we want to close the channels???
+				// close(t.Ch)
+				// close(t.input)
+				return
+			}
+		case <-t.timer.C:
+			t.isSet = false
+			t.Ch <- struct{}{}
```
ebuchman: this was non-blocking before. I see it's buffered (cap == 1) but otherwise this is now blocking. any particular reason not to keep that old behaviour?

common/throttle_timer.go
```diff
@@ -12,54 +11,85 @@ If a long continuous burst of .Set() calls happens, ThrottleTimer fires
...
-	case <-t.quit:
-		// do nothing
+func (t *ThrottleTimer) run() {
+	for {
+		select {
+		case cmd := <-t.input:
+			// stop goroutine if the input says so
+			if t.processInput(cmd) {
+				// TODO: do we want to close the channels???
+				// close(t.Ch)
```
melekes: I think we should close Ch to indicate there will be no more ticks. no need to close internal `input` channel (it will be closed automatically)

common/throttle_timer.go
```diff
@@ -12,54 +11,85 @@ If a long continuous burst of .Set() calls happens, ThrottleTimer fires
...
+func (t *ThrottleTimer) processInput(cmd command) (shutdown bool) {
+	switch cmd {
+	case Set:
+		if !t.isSet {
+			t.isSet = true
+			t.timer.Reset(t.dur)
+		}
+	case Quit:
+		shutdown = true
+		fallthrough
```
melekes: why `fallthough`?

common/throttle_timer.go
```diff
@@ -12,54 +11,85 @@ If a long continuous burst of .Set() calls happens, ThrottleTimer fires
...
+	case Unset:
+		if t.isSet {
+			t.isSet = false
+			t.timer.Stop()
+		}
 	default:
-		t.timer.Reset(t.dur)
+		panic("unknown command!")
 	}
+	// return true
```
melekes: huh?

ethanfrey: sorry test code

common/throttle_timer.go
ethanfrey: quit also wants to stop any running timer, like unset

common/throttle_timer.go
ethanfrey: :+1:

common/throttle_timer.go
ethanfrey: Thank you. Very good.

common/throttle_timer.go
ethanfrey: In the old model, if it was not read immediately, the message got dropped, and effectively called `Set` again to schedule it for later.

This model should keep the number of ticks received equal to the number produced. The timer blocks, sure, but is there a downside to that?

Hmmm... I just checked how `time` does it, seem like they prefer to drop as you did.
https://golang.org/src/time/tick.go#L16
https://golang.org/src/time/sleep.go#L132

I'll update to that.

common/throttle_timer.go
ethanfrey: Not so fast...

```
./throttle_timer.go:50:10: invalid operation: close(t.Ch) (cannot close receive-only channel)
./throttle_timer.go:55:9: invalid operation: t.Ch <- struct {} literal (send to receive-only type <-chan struct {})
./throttle_timer_test.go:49:23: cannot use t.Ch (type <-chan struct {}) as type chan struct {} in field value
```

I think I need to expose it read only and keep a write copy as a private variable?

ethanfrey:
Okay, I tried to integrate feedback, I would love more feedback.
After reading your non-blocking comment and looking at the implementation of time in stdlib, I submitted https://github.com/tendermint/tmlibs/pull/94/commits/1ac4c5dd6d007a708337e1ad2636e456e1e4b8db

Is this good?  Is there a better/cleaner way you can suggest?

common/throttle_timer.go
melekes: yeah, this was a bad idea given current impl. If we want to make this right, we need to unexpose Ch and create a method >>

```
func (t ...) Ch() <-chan struct{} {
  return t.Ch
}

```

ebuchman: Ah yes. The right way would be to make this private and have a method that returns the `<-chan`. But that would make this breaking. Why don't we expose that method and mark the `Ch` as deprecated, and make private at some future release. keeping around the `output` copy is weird, let's drop it

melekes: I don't mind breaking changes

jaekwon:
> As I was reviewing abci, I came across throttle timer, which seemed very confusing code to me, in that it mixed mutex and channel, and accessed state across multiple goroutines (some generated internally to the time package via AfterFunc). I added tests and proved to myself it worked properly, but I wanted to try my hand at refactoring it. One for cleaner code, and also for me to get more comfortable with how goroutines and channels are used in tendermint code.

We just have to get used to mixing mutex locks, waitgroups, channels, and goroutines.  And after that, try waitcond.

It's not hard to use locks when you use the right pattern.  In the previous code, we had a code segment like this:
```golang
type ThrottleTimer struct {
        ...

	mtx   sync.Mutex		
  	timer *time.Timer		  	timer *time.Timer
  	isSet bool
}
```

This is convention that is used all over the Tendermint/Cosmos codebases where 'timer' and 'isSet' are both protected under the mutex lock above.  Basically, any time you change those values, you must lock first and unlock after.  Because all you're doing is reassining those values, there's typically no reason that any caller would have to block too long... in fact, using this convention is faster than using channels because channels require more than one lock to implement.  And there's actually less code and control flow.  The tricky part is that you need to put the locks and unlocks in the right places, but that gets easier with practice.  It's a good idea to almost always use Lock() and defer Unlock() right after... and if this is bad because you need to unlock before performing more (blocking/slow) logic, then you probably just need to refactor your code to move what needs to happen between Lock and Unlock (inclusive) in a new function (while keeping the defer statement).

Some other conventions and rules off the top of my head:

* When naming methods that are supposed to run in separate goroutines, name them "XYZRoutine".
* Do not write logic that requires the caller to drain channels.  It might create zombie goroutines that aren't garbage collected, and it's an example of bleeding abstractions.
* Don't mix sender routines and receiver routines.  There should generally be a single goroutine that sends to a channel, and *it* is responsible for closing the channel, nobody else.
* Sometimes what you want is an RWMutex, WaitGroup, or Cond.  If you use something else where you are supposed to use these, you will probably have a bad time.

Lets revert these changes and adopt the policy of mixing locks/goroutines/channels/waitgroups where appropriate, and start adopting a uniform convention for using them.  Take a look at the Tendermint codebase for the existing conventions.

PR #95, opened 2017-12-07
Refactor repeat timer

ethanfrey:
Same cleanup (remove mutex, use command channel) as with throttle timer.
Made the output non-blocking as per time.Ticker
Made Stop idemnepotent

### DIFF
--- a/common/repeat_timer.go
+++ b/common/repeat_timer.go
@@ -1,7 +1,7 @@
 package common
 
 import (
-	"sync"
+	"fmt"
 	"time"
 )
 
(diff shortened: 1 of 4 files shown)


common/repeat_timer.go
```diff
@@ -11,76 +10,93 @@ It's good for keeping connections alive.
...
-	ticker *time.Ticker
-	quit   chan struct{}
-	wg     *sync.WaitGroup
-	dur    time.Duration
+	dur     time.Duration
+	ticker  *time.Ticker
+	stopped bool
 }
 
+type repeatCommand int32
```
melekes: why int32?) int8 would cut it, no

common/repeat_timer.go
```diff
@@ -11,76 +10,93 @@ It's good for keeping connections alive.
...
-		close(t.quit)
-		t.wg.Wait() // must wait for quit to close else we race Reset
-		t.ticker = nil
 	}
-	return exists
+}
+
+// trySend performs non-blocking send on t.Ch
+func (t *RepeatTimer) trySend() {
+	// NOTE: this was blocking in previous version (t.Ch <- t_)
```
melekes: we'll need to test Tendermint (or where this timer is used) if we're gonna change behaviour

common/repeat_timer.go
ethanfrey: yup

common/repeat_timer.go
ethanfrey: will return to old behavior and leave comment for now

ethanfrey:
Okay, anton, take another look if you want

ethanfrey:
Thanks for the cleanup and merge!

common/repeat_timer.go
```diff
@@ -11,76 +10,93 @@ It's good for keeping connections alive.
...
 		}
-		close(t.quit)
-		t.wg.Wait() // must wait for quit to close else we race Reset
-		t.ticker = nil
 	}
-	return exists
+}
+
+// send performs blocking send on t.Ch
+func (t *RepeatTimer) send(tick time.Time) {
```
ebuchman: we should defntly not block! I think this is responsible for some of the recent non-deterministic failures on Tendermint. I'll change this

PR #98, opened 2017-12-13
Sdk2 iterator

jaekwon:
See https://github.com/cosmos/cosmos-sdk/pull/303


db/backend_test.go
```diff
@@ -2,42 +2,84 @@ package db
...
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	cmn "github.com/tendermint/tmlibs/common"
 )
 
-func testBackend(t *testing.T, backend string) {
+func cleanupDBDir(dir, name string) {
+	os.RemoveAll(path.Join(dir, name) + ".db")
```
melekes: `filepath` please (path does not work on Windows)

db/backend_test.go
```diff
@@ -2,42 +2,84 @@ package db
...
-	testBackend(t, CLevelDBBackendStr)
-	testBackend(t, GoLevelDBBackendStr)
-	testBackend(t, MemDBBackendStr)
+func TestBackendsGetSetDelete(t *testing.T) {
+	for dbType, _ := range backends {
+		testBackendGetSetDelete(t, dbType)
+	}
+}
+
+func assertPanics(t *testing.T, dbType, name string, fn func()) {
```
melekes: https://github.com/stretchr/testify/blob/2aa2c176b9dab406a6970f6a55f513e8a8c8b18f/assert/assertions.go#L798

db/fsdb.go
```diff
@@ -59,22 +59,40 @@ func (db *FSDB) Get(key []byte) []byte {
...
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
+	panicNilKey(key)
+
+	path := db.nameToPath(key)
+	_, err := read(path)
+	if os.IsNotExist(err) {
+		return false
+	} else if err != nil {
+		panic(errors.Wrap(err, fmt.Sprintf("Getting key %s (0x%X)", string(key), key)))
```
melekes: https://github.com/pkg/errors/blob/8842a6e0cc595d1cc9d931f6c875883967280e32/errors.go#L194 `errors.Wrapf`

db/go_level_db.go
```diff
@@ -156,17 +169,31 @@ func (mBatch *goLevelDBBatch) Write() {
...
 
-func (db *GoLevelDB) Iterator() Iterator {
+func (db *GoLevelDB) Iterator(start, end []byte) Iterator {
 	itr := &goLevelDBIterator{
 		source: db.db.NewIterator(nil, nil),
+		start:  start,
+		end:    end,
 	}
-	itr.Seek(nil)
+	itr.source.Seek(start)
```
melekes: what about the same check above?

```
if len(start) > 0 {
  itr.source.Seek(start)
} else {
  itr.source.SeekToFirst()
}
```

db/types.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// For testing convenience.
+func bz(s string) []byte {
+	return []byte(s)
+}
+
+// All DB funcs should panic on nil key.
+func panicNilKey(key []byte) {
+	if key == nil {
```
melekes: should we also check `len(key) == 0` here?

melekes:
```
db/stats.go:3:1:warning: mergeStats is unused (deadcode)
clist/clist.go:96::warning: Use of unsafe calls should be audited,LOW,HIGH (gas)
clist/clist.go:112::warning: Use of unsafe calls should be audited,LOW,HIGH (gas)
common/os.go:27::warning: Subprocess launching with partial path.,MEDIUM,HIGH (gas)
common/random.go:60::warning: Use of weak random number generator (math/rand instead of crypto/rand),HIGH,MEDIUM (gas)
common/random.go:76::warning: Use of weak random number generator (math/rand instead of crypto/rand),HIGH,MEDIUM (gas)
process/process.go:28::warning: Subprocess launching with variable.,HIGH,HIGH (gas)
db/stats.go:3:6:warning: func mergeStats is unused (U1000) (megacheck)
```

I thought we disabled this thing ... strange (cc @zramsay )

db/backend_test.go
ethanfrey: :+1:

db/go_level_db.go
ebuchman: C-LevelDB requires that the key itself is not empty, though it's not documented, so you can't pass an empty key: https://github.com/jmhodges/levigo/blob/master/iterator.go#L127

GoLevelDB seems to be fine with empty keys. But you're right, would probably be better to check.

db/backend_test.go
```diff
@@ -2,42 +2,76 @@ package db
...
+func cleanupDBDir(dir, name string) {
+	os.RemoveAll(filepath.Join(dir, name) + ".db")
+}
+
+func testBackendGetSetDelete(t *testing.T, backend string) {
 	// Default
 	dir, dirname := cmn.Tempdir(fmt.Sprintf("test_backend_%s_", backend))
 	defer dir.Close()
 	db := NewDB("testdb", backend, dirname)
-	require.Nil(t, db.Get([]byte("")))
```
odeke-em: I believe []byte("") and nil are pretty essential keys for unit tests, as they are the blank keys and defaults usually so can help weed out any stale internals.

db/types.go
```diff
@@ -0,0 +1,124 @@
...
+	ReverseIterator(start, end []byte) Iterator
+
+	// Releases the connection.
+	Close()
+
+	// Creates a batch for atomic updates.
+	NewBatch() Batch
+
+	// For debugging
+	Print()
```
odeke-em: Personally I'd prefer
```go
Fprintf(w io.Writer)
```

Otherwise our stdout is about to get lit for every database and of which extending that to print to the screen during tests will just be a matter of `Fprint(os.Stdout)`. Using `Fprint` here all of a sudden gives us the power
to test our DBs easily, periodically print out output to a log file for consistency checks, send overview over a network etc.

db/types.go
```diff
@@ -0,0 +1,124 @@
...
+	// Stats returns a map of property values for all keys and the size of the cache.
+	Stats() map[string]string
+}
+
+//----------------------------------------
+// Batch
+
+type Batch interface {
+	SetDeleter
+	Write()
```
odeke-em: No error?

db/types.go
```diff
@@ -0,0 +1,124 @@
...
+	// If Valid returns false, this method will panic.
+	Key() []byte
+
+	// Value returns the value of the cursor.
+	//
+	// If Valid returns false, this method will panic.
+	Value() []byte
+
+	// Release deallocates the given Iterator.
+	Release()
```
odeke-em: Yeah am not too sold on Release(), personally I'd go for `Close() error`

db/types.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// For testing convenience.
+func bz(s string) []byte {
+	return []byte(s)
+}
+
+// All DB funcs should panic on nil key.
```
odeke-em: This seems like something that might come to bite us later on and might add a mental overload
since now developers will have to read the docs to know that only our DBs panic when they
get a nil key, of which if am layers deep in code, at runtime having to check if a key is non-nil first
before a lookup might be asking for too much.

db/util.go
```diff
@@ -1,82 +1,46 @@
...
-
-func IteratePrefix(db DB, prefix []byte) Iterator {
-	itr := db.Iterator()
-	pi := &prefixIterator{
-		itr:    itr,
-		prefix: prefix,
-	}
-	pi.Seek(prefix)
-	return pi
+func IsKeyInDomain(key string, start, end []byte) bool {
```
odeke-em: Could we change this signature to take key in as a []byte. Notice that in every call here we are incuring
O(2*n) []byte<-->string conversions whereas by making the signature take in `key []byte`, we don't have to do any strings.Compare and string(start), string(end) comparisons. bytes already has bytes.Compare.

db/backend_test.go
jaekwon: yeah makes sense to add this back

db/backend_test.go
```diff
@@ -2,42 +2,76 @@ package db
...
-func TestBackends(t *testing.T) {
-	testBackend(t, CLevelDBBackendStr)
-	testBackend(t, GoLevelDBBackendStr)
-	testBackend(t, MemDBBackendStr)
+func TestBackendsNilKeys(t *testing.T) {
+	// test all backends
+	for dbType, creator := range backends {
+		name := cmn.Fmt("test_%x", cmn.RandStr(12))
+		db, err := creator(name, "")
+		defer cleanupDBDir("", name)
```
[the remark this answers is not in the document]
jaekwon: seems extremely unlikely.  we should move db.Close() up as a defer.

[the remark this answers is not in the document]
jaekwon: yeah.  defer doesn't work on for-loops, only function boundaries.

probably better to create a function with defer statements, and call that function in the for loop.

db/backend_test.go
```diff
@@ -2,42 +2,76 @@ package db
...
+		db.Close()
+	}
+}
+
+func TestGoLevelDBBackendStr(t *testing.T) {
+	name := cmn.Fmt("test_%x", cmn.RandStr(12))
+	db := NewDB(name, LevelDBBackendStr, "")
+	defer cleanupDBDir("", name)
+
+	if _, ok := backends[CLevelDBBackendStr]; !ok {
```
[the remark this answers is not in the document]
jaekwon: yeah it's silly.  We should remove this if statement and just get GoLevelDBBackendStr explicitly, db := NewDB(name, GoLevelDBBackendStr, "")

db/c_level_db.go
```diff
@@ -17,17 +18,17 @@ func init() {
...
 
+var _ DB = (*CLevelDB)(nil)
+
 type CLevelDB struct {
 	db     *levigo.DB
 	ro     *levigo.ReadOptions
 	wo     *levigo.WriteOptions
 	woSync *levigo.WriteOptions
-
-	cwwMutex
```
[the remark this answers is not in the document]
jaekwon: Yeah.  No point making magic if it doesn't go all the way to make it safe.

db/c_level_db.go
```diff
@@ -45,42 +46,50 @@ func NewCLevelDB(name string, dir string) (*CLevelDB, error) {
...
 		wo:     wo,
 		woSync: woSync,
-
-		cwwMutex: NewCWWMutex(),
 	}
 	return database, nil
 }
 
 func (db *CLevelDB) Get(key []byte) []byte {
+	panicNilKey(key)
```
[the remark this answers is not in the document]
jaekwon: No, this is the spec.  You can't fetch a nil key.  I suspect it'll help people catch some errors.

db/c_level_db.go
```diff
@@ -45,42 +46,50 @@ func NewCLevelDB(name string, dir string) (*CLevelDB, error) {
...
 	return res
 }
 
+func (db *CLevelDB) Has(key []byte) bool {
+	panicNilKey(key)
+	panic("not implemented yet")
+}
+
 func (db *CLevelDB) Set(key []byte, value []byte) {
+	panicNilKey(key)
```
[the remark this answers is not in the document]
jaekwon: see other comment

db/c_level_db.go
```diff
@@ -155,59 +160,83 @@ func (mBatch *cLevelDBBatch) Write() {
 //----------------------------------------
 // Iterator
 
-func (db *CLevelDB) Iterator() Iterator {
+func (db *CLevelDB) Iterator(start, end []byte) Iterator {
```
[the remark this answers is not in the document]
jaekwon: Seek/Prev complicates logic (e.g. will ripple through to the SDK store module).

Seek() isn't used anywhere. It was used in tests because it was the only way to seek.  Now we have start/end instead.

db/c_level_db.go
```diff
@@ -155,59 +160,83 @@ func (mBatch *cLevelDBBatch) Write() {
...
+	key := c.itr.Key()
+	if c.end != nil && bytes.Compare(key, c.end) > 0 {
+		c.invalid = true
 	}
-	c.itr.Prev()
+	return key
 }
 
-func (c cLevelDBIterator) Close() {
+func (c *cLevelDBIterator) Release() {
```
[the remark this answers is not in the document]
jaekwon: Close() works!

db/fsdb.go
```diff
@@ -39,73 +38,95 @@ func NewFSDB(dir string) *FSDB {
...
-		cwwMutex: NewCWWMutex(),
+		dir: dir,
 	}
 	return database
 }
 
 func (db *FSDB) Get(key []byte) []byte {
 	db.mtx.Lock()
 	defer db.mtx.Unlock()
+	panicNilKey(key)
```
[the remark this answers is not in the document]
jaekwon: If we allow setting nil, we shouldn't differentiate between nil and the empty byte slice, but then what gets set and what gets returned are different.  Better to force consistency from the getgo.

Also Iterator(start, end) nil 'end' has a special meaning, so the two reasons work together.

db/fsdb.go
```diff
@@ -39,73 +38,95 @@ func NewFSDB(dir string) *FSDB {
...
 	return value
 }
 
+func (db *FSDB) Has(key []byte) bool {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
+	panicNilKey(key)
+
+	path := db.nameToPath(key)
+	_, err := read(path)
```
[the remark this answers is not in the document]
jaekwon: great.

db/types.go
```diff
@@ -0,0 +1,124 @@
...
+	// CONTRACT: No writes may happen within a domain while an iterator exists over it.
+	Iterator(start, end []byte) Iterator
+
+	// Iterator over a domain of keys in descending order. End is exclusive.
+	// Start must be greater than end, or the Iterator is invalid.
+	// CONTRACT: No writes may happen within a domain while an iterator exists over it.
+	ReverseIterator(start, end []byte) Iterator
+
+	// Releases the connection.
+	Close()
```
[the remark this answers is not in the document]
jaekwon: no, we should panic on all errors because there's no way to safely handle nondeterministic failures meaningfully.  a panic trace provides more information than an error.

just restart your node after fixing the problem and sync again.

db/types.go
jaekwon: Lets make another PR for that.

jaekwon: We shouldn't be using Print for consistency checks (it's explicitly for human consumption) but it makes sense to create an overview to send over the network (e.g. RPC endpoint).  I'd prefer Sprint() string though.

jaekwon: That way we kinda implicitly tell the user that this isn't designed for large outputs., just for debugging.

db/types.go
jaekwon: No error for any IO operation.  See other comment.

db/types.go
jaekwon: Close is fine.

db/types.go
jaekwon: see other comment.

db/types.go
jaekwon: no, this function specifically checks for NilKey.

db/util.go
jaekwon: yeah it should be []byte.

db/c_level_db.go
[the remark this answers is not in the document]
jaekwon: I wasn't, but it makes sense to change the spec to allow nil keys.
It'll probably prevent more bugs than it creates.  Thanks!

db/c_level_db.go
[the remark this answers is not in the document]
odeke-em: I see I see, thanks for pointing that out.

db/types.go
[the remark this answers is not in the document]
odeke-em: That's a valid point, and yes it applies for our tendermint CLI. However, IMHO if we are making an SDK, for users that aren't going to necessarily make CLI apps panicking is going to trip them out. If I put this in a webapp or even a UI app, I would rather handle the error than letting my app crash.

db/types.go
odeke-em: Gotcha, I read your other comment, thanks.

db/types.go
odeke-em: Thanks!

odeke-em:
@jaekwon I've pushed up some of the changes that we agreed on in the comments above, PTAL and thank you!

db/backend_test.go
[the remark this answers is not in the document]
ebuchman: the point of this was just to test what the default `LevelDB` is. We can remove it.

db/types.go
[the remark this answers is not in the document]
ebuchman: The SDK is specifically for building deterministic state machines.  If there's a problem with your database, it's not an error you're going to handle in your app logic. We can provide a recover function somewhere at the top of the sdk execution context that will return errors and tell them their code or devices are broken (with stack trace)

ebuchman:
ok this LGTM

Issue #99, opened 2017-12-13
rand: perhaps make an internal rand.RandSource that's concurrent safe to avoid nefarious rand.Seed(1) resets


odeke-em:
### Problem Statement
As a followup to https://github.com/tendermint/tendermint/issues/973, inside some Tendermint repos, we'll see direct invocations of math/rand.R* such as here in:
https://github.com/tendermint/tmlibs/blob/bfcc0217f120d3bee6730ba0789d2eb72fc2e889/common/bit_array.go#L215
https://github.com/tendermint/tmlibs/blob/master/common/bit_array.go#L220
https://github.com/tendermint/tmlibs/blob/master/common/bit_array.go#L235

where we trustingly assumed that at init we properly seeded the random source with 
https://github.com/tendermint/tmlibs/blob/bfcc0217f120d3bee6730ba0789d2eb72fc2e889/common/random.go#L13-L21
but as shown in issue https://github.com/tendermint/tendermint/issues/973, a couple of our dependencies also set `rand.Seed` hence if our init gets invoked before theirs, all our initial seeding is no longer useful

### Analysis
I think our usages of math/rand.R* are predicated on the ease of use of the stdlib and the fact that the default source guarantees to be concurrency safe but others don't https://golang.org/pkg/math/rand/
<img width="937" alt="screen shot 2017-12-13 at 11 20 02 am" src="https://user-images.githubusercontent.com/4898263/33955086-99b52282-dff7-11e7-97a2-dce2497ca312.png">

### Proposed fix
Mirror the functionality of the default source with our own internal RNG seeded with crypto/rand's true randomness and then gut/replace every usage of math/rand in all Tendermint's code

PR #100, opened 2017-12-14
common: no more relying on math/rand.DefaultSource

odeke-em:
Fixes https://github.com/tendermint/tmlibs/issues/99
Updates https://github.com/tendermint/tendermint/issues/973

Removed usages of math/rand.DefaultSource in favour of our
own source that's seeded with a completely random source
and is safe for use in concurrent in multiple goroutines.
Also extend some functionality that the stdlib exposes such as
* RandPerm
* RandIntn
* RandInt31
* RandInt63

Also added an integration test whose purpose is to be run as
a consistency check to ensure that our results never repeat
hence that our internal PRNG is uniquely seeded each time.

### DIFF
--- a/common/random.go
+++ b/common/random.go
@@ -35,6 +35,7 @@ func init() {
 }
 
 // Constructs an alphanumeric string of given length.
+// It is not safe for cryptographic usage.
 func RandStr(length int) string {
 	chars := []byte{}
 MAIN_LOOP:
@@ -58,10 +59,12 @@ MAIN_LOOP:
 	return string(chars)
 }
 
+// It is not safe for cryptographic usage.
 func RandUint16() uint16 {
 	return uint16(RandUint32() & (1<<16 - 1))
 }
 
+// It is not safe for cryptographic usage.
 func RandUint32() uint32 {
 	prng.Lock()
 	u32 := prng.Uint32()
@@ -69,10 +72,12 @@ func RandUint32() uint32 {
 	return u32
 }
 
+// It is not safe for cryptographic usage.
 func RandUint64() uint64 {
 	return uint64(RandUint32())<<32 + uint64(RandUint32())
 }
 
+// It is not safe for cryptographic usage.
 func RandUint() uint {
 	prng.Lock()
 	i := prng.Int()
@@ -80,18 +85,22 @@ func RandUint() uint {
 	return uint(i)
 }
 
+// It is not safe for cryptographic usage.
 func RandInt16() int16 {
 	return int16(RandUint32() & (1<<16 - 1))
 }
 
+// It is not safe for cryptographic usage.
 func RandInt32() int32 {
 	return int32(RandUint32())
 }
 
+// It is not safe for cryptographic usage.
 func RandInt64() int64 {
 	return int64(RandUint64())
 }
 
+// It is not safe for cryptographic usage.
 func RandInt() int {
 	prng.Lock()
 	i := prng.Int()
@@ -99,6 +108,7 @@ func RandInt() int {
 	return i
 }
 
+// It is not safe for cryptographic usage.
 func RandInt31() int32 {
 	prng.Lock()
 	i31 := prng.Int31()
@@ -106,6 +116,7 @@ func RandInt31() int32 {
 	return i31
 }
 
+// It is not safe for cryptographic usage.
 func RandInt63() int64 {
 	prng.Lock()
 	i63 := prng.Int63()
@@ -114,6 +125,7 @@ func RandInt63() int64 {
 }
 
 // Distributed pseudo-exponentially to test for various cases
+// It is not safe for cryptographic usage.
 func RandUint16Exp() uint16 {
 	bits := RandUint32() % 16
 	if bits == 0 {
@@ -125,6 +137,7 @@ func RandUint16Exp() uint16 {
 }
 
 // Distributed pseudo-exponentially to test for various cases
+// It is not safe for cryptographic usage.
 func RandUint32Exp() uint32 {
 	bits := RandUint32() % 32
 	if bits == 0 {
@@ -136,6 +149,7 @@ func RandUint32Exp() uint32 {
 }
 
 // Distributed pseudo-exponentially to test for various cases
+// It is not safe for cryptographic usage.
 func RandUint64Exp() uint64 {
 	bits := RandUint32() % 64
 	if bits == 0 {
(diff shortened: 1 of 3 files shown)


common/random_test.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+// It is essential that these tests run and never repeat their outputs
+// lest we've been pwned and the behavior of our randomness is controlled.
+// See Issues:
+//  * https://github.com/tendermint/tmlibs/issues/99
+//  * https://github.com/tendermint/tendermint/issues/973
+func TestUniqueRng(t *testing.T) {
+	if os.Getenv("TENDERMINT_INTEGRATION_TESTS") == "" {
+		t.Skipf("Can only be run as an integration test")
```
melekes: why, too time consuming?

ebuchman:
Pushed a commit.

Re the crazy test. I see a couple of possible improvements:

1) introduce a `reset()` and just call this from `init()`. Then we can call `reset()` from the test and run the whole thing in-process
2) if we don't like (1), can we use `go build` to build the binary once and then call it in the loop instead of `go run` every time. I think that would speed it up. Would be better than the integration variable

common/random_test.go
odeke-em: In deed, yes it takes about 54s to run altogether.

odeke-em:
Thank you @ebuchman for the commit and for the thoughts.
1. I like your idea, simple and pragmatic. The only problem though is that it relies on invoking `init()` which by paranoia is the whole point of that integration test: external code run through a normal import. However, it solves the problem of wonky environments and $GOPATH not being properly setup in tests, plus it is quite fast.
2. I had actually thought of doing this, but then we have to write special handling for Windows.

I like 1 the most and I'll push up something for it, thanks again.

common/random.go
```diff
@@ -45,84 +51,123 @@ MAIN_LOOP:
...
 	return time.Unix(int64(RandUint64Exp()), 0)
 }
 
 func RandBytes(n int) []byte {
-	bs := make([]byte, n)
-	for i := 0; i < n; i++ {
-		bs[i] = byte(rand.Intn(256))
-	}
-	return bs
+	return cRandBytes(n)
```
jaekwon: We should keep the old one... cRandBytes is not guaranteed to be fast...

jaekwon: See PR comments

odeke-em: Gotcha gotcha, that's reasonable.

jaekwon:
Lets add comments on every Rand*() that it's not safe for cryptographic use.
For all randomness that must be cryptographically safe, you must use CRand*().

The benefit of Rand*() is that it's guaranteed to be fast.
(see my comment about RandBytes not using cRandBytes).
Arguably CRand() should still be fast for the same reason that /dev/urandom is fast, but it's good not to trust anything, and the only provably good way to get randomness is to get it from quantum effects, and that may slow down some implementations of CRand*.

So the difference between Rand and CRand is the same as what people *used to* think was the difference between /dev/urandom and /dev/random.

https://www.2uo.de/myths-about-urandom/

odeke-em:
@jaekwon so our PRNG is seeded with CRand* and that's the entire impetus of making this PR to avoid a controllable PRNG that's also safe for cryptographic use. That was our initial promise but then we noticed that we used rand.Seed in there after reading CRand* and then using math/rand.
/cc @ebuchman

jaekwon:
Yeah but the contract is that Rand() isn't safe for cryptographic use.  We can try to make it safe, but it isn't and we should warn users to not use it that way.

Rand() is our helper functions for math/rand.
CRand() is our helper functions for crypto/rand.

jaekwon:
That said, this PR is good because it fixes a bug whereby Rand* is sometimes not random.

odeke-em:
Roger that @jaekwon, I agree with your wording since we use the PRNG from math. However, I am conflicted because we seed it entirely from `crypto/rand`. Could you help me with some ideas on the wording for this? I mention this because all our code uses functions from here, assuming that these funcs are safe for cryptographic usage.

common/random.go
odeke-em: Oh man, great call @jaekwon to ask to revert this. I just created and ran benchmarks and using cRandBytes(n) is between ~3.5X slower!!

odeke-em: ### Before this change with cRand
```shell
$ go test -run=^$ -bench=.
goos: darwin
goarch: amd64
pkg: github.com/tendermint/tmlibs/common
BenchmarkRandBytes10B-4      	 5000000	       276 ns/op
BenchmarkRandBytes100B-4     	  500000	      2542 ns/op
BenchmarkRandBytes1KiB-4     	   50000	     27274 ns/op
BenchmarkRandBytes100KiB-4   	      50	  27388692 ns/op
BenchmarkRandBytes1MiB-4     	       1	29350694138 ns/op
PASS
```

### After this change
```shell
ok  	github.com/tendermint/tmlibs/common	35.452s
Emmanuels-MacBook-Pro-2:common emmanuelodeke$ go test -run=^$ -bench=.
goos: darwin
goarch: amd64
pkg: github.com/tendermint/tmlibs/common
BenchmarkRandBytes10B-4      	 2000000	       806 ns/op
BenchmarkRandBytes100B-4     	  200000	      7210 ns/op
BenchmarkRandBytes1KiB-4     	   20000	     73042 ns/op
BenchmarkRandBytes100KiB-4   	      20	  79803972 ns/op
BenchmarkRandBytes1MiB-4     	       1	81293929913 ns/op
PASS
ok  	github.com/tendermint/tmlibs/common	89.242s
```
yeezy yeezy!!

jaekwon:
>  I mention this because all our code uses functions from here, assuming that these funcs are safe for cryptographic usage.

They do not, and should not.

We should explicitly say that these are not safe for cryptographic use.  Even math/rand explicitly says not to use it for cryptographic use.

jaekwon:
Rand() and math/rand are trivial not secure.

`mrand.New(mrand.NewSource(int64(seed)))`.  Int64, 64 bits of randomness.

jaekwon:
The general rule of thumb used to be that you need at least 80 bits of entropy to be secure.

For example, the last Bitcoin blockchain hash I saw had 73 leading binary 0's.

odeke-em:
That was an awesome discussion and lesson that we had offline, thank you @jaekwon!

common/random.go
```diff
@@ -2,30 +2,45 @@ package common
...
 	"time"
 )
 
 const (
 	strChars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz" // 62 characters
 )
 
-func init() {
+// pseudo random number generator.
+// seeded with OS randomness (crand)
```
melekes: `mrand`, not `crand` ??

odeke-em: Nope, it is seeded with crand, see init()

PR #101, opened 2017-12-17
Nil keys are OK, deprecate BeginningKey/EndingKey

jaekwon:
Before we didn't allow nil keys, and there was a concept of BeginningKey (empty) and EndingKey (nil) to iterate/reverse-iterate over ranges.

The previous API was bad for several reasons:
* most Golang APIs generally let nil and empty byteslices be interchangeable
* the new DB API is more intuitive, and doesn't require BeginningKey/EndingKey for iteration.  See documentation on db/types.DB for more details.


db/backend_test.go
```diff
@@ -55,27 +62,89 @@ func withDB(t *testing.T, creator dbCreator, fn func(DB)) {
...
+
+				// Set empty, Delete nil
+				db.Set([]byte(""), []byte("abc"))
+				expect(nil, []byte("abc"))
+				db.Delete(nil)
+				expect(nil, nil)
+
+				// Set empty, Delete empty
+				db.Set([]byte(""), []byte("abc"))
+				expect(nil, []byte("abc"))
```
melekes: minor: should this be `expect([]byte(""), []byte("abc"))`. i understand that nil and []byte("") are interchangeble

db/go_level_db.go
```diff
@@ -169,102 +162,104 @@ func (mBatch *goLevelDBBatch) Write() {
...
-	if len(start) == 0 {
-		start = nil
+func newGoLevelDBIterator(source iterator.Iterator, start, end []byte, isReverse bool) *goLevelDBIterator {
+	if isReverse {
+		panic("not implemented yet") // XXX
 	}
-	return &util.Range{
-		Start: start,
-		Limit: end,
+	source.Seek(start)
```
melekes: shouldn't we check for nil?

db/go_level_db.go
```diff
@@ -169,102 +162,104 @@ func (mBatch *goLevelDBBatch) Write() {
...
+	return &goLevelDBIterator{
+		source:    source,
+		start:     start,
+		end:       end,
+		isReverse: false,
```
melekes: should be set to `isReverse`

db/util.go
```diff
@@ -35,11 +35,26 @@ func cpIncr(bz []byte) (ret []byte) {
...
+		}
+		if end != nil && bytes.Compare(end, key) <= 0 {
+			return false
+		}
+		return true
+	} else {
+		if start != nil && bytes.Compare(start, key) < 0 {
+			return false
+		}
+		if end != nil && bytes.Compare(key, end) <= 0 {
```
odeke-em: Won't this condition always return false say
given:
key:   []byte("c")
start: []byte("a")
end:   []byte("z")
reverse: false

Given the docs at https://github.com/tendermint/tmlibs/blob/091b416370d448f03390c68dc2f663614cbaac36/db/types.go#L23-L28

odeke-em: Or perhaps could you explain to me what IsKeyInDomain is supposed to do?

db/types.go
```diff
@@ -116,9 +114,12 @@ func bz(s string) []byte {
 	return []byte(s)
 }
 
-// All DB funcs should panic on nil key.
-func panicNilKey(key []byte) {
-	if key == nil {
-		panic("nil key")
+// We defensively turn nil keys or values into []byte{} for
+// most operations.
+func nonNilBytes(bz []byte) []byte {
```
odeke-em: Nice, this is more comforting to use than us panicking on every nil encounter :)

db/go_level_db.go
```diff
@@ -169,102 +162,104 @@ func (mBatch *goLevelDBBatch) Write() {
...
-	v := make([]byte, len(val))
-	copy(v, val)
-
-	return v
-}
 
-func (it *goLevelDBIterator) Valid() bool {
-	it.assertNoError()
-	if it.invalid {
+	// If key is end or past it, invalid.
```
ebuchman: technically we don't need to do this here since go-leveldb lets you specify the range in the iterator.

but for consistency with c-leveldb, ok

db/util.go
```diff
@@ -35,11 +35,26 @@ func cpIncr(bz []byte) (ret []byte) {
...
 }
 
-func IsKeyInDomain(key, start, end []byte) bool {
-	leftCondition := bytes.Equal(start, BeginningKey()) || bytes.Compare(key, start) >= 0
-	rightCondition := bytes.Equal(end, EndingKey()) || bytes.Compare(key, end) < 0
-	return leftCondition && rightCondition
+// See DB interface documentation for more information.
+func IsKeyInDomain(key, start, end []byte, isReverse bool) bool {
+	if !isReverse {
+		if bytes.Compare(key, start) < 0 {
```
ebuchman: shouldnt we have `start != nil` here too ?

db/util.go
ebuchman: if `isReverse==false` we're in the first branch,  so in the example you've given we'd return `true`

db/backend_test.go
jaekwon: that's why it should be nil, cuz it's easier to type :P

db/go_level_db.go
jaekwon: I don't think so, otherwise the tests would have failed.  It did fail for cleveldb and I fixed it.

db/util.go
jaekwon: They're the same: https://play.golang.org/p/prE7JB_tTX

db/go_level_db.go
jaekwon: ok

PR #107, opened 2017-12-21
Fix/cmn

ebuchman:
Fixes a race in random.go and broken tests for repeat timer

### DIFF
--- a/common/random.go
+++ b/common/random.go
@@ -40,7 +40,7 @@ func RandStr(length int) string {
 	chars := []byte{}
 MAIN_LOOP:
 	for {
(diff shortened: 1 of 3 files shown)


odeke-em:
What was the data race @ebuchman? How did you find/reproduce it?

ebuchman:
Tendermint test failures. 

https://github.com/tendermint/tmlibs/pull/107/files#diff-74fa9a04cef91879f95798811e00671dL43 is clearly racey if since it doesn't grab the prng lock.

ticker had broken tests since it depended on sleeps so I introduced the ManualTicker to fix

PR #109, opened 2017-12-22
Rational

rigelrozanski:
Basic fraction functionality, used for computations within gaia for staking and various other crypto-economic mechanisms which require fraction-precision.

Note that Evaluate() uses bankers rounding!


melekes:
```
fraction/fraction.go:63:5:warning: ineffective assignment to field Numerator (SA4005) (megacheck)
```

melekes:
Could you fill in the description? I can merge np, but still interested to know where we're using it.

melekes:
Reported crash: https://allinbits.slack.com/files/U833P655L/F8JLN9BB6/gaia_crash.go

rigelrozanski:
@melekes crash report not relevant - as it's not using this `tmlibs/fraction`!  That gaia branch needs to be updated... just wanted this PR here for reference or if you have any other architectural recommendations..... 

> fraction/fraction.go:63:5:warning: ineffective assignment to field Numerator (SA4005) (megacheck)

what did you run to get this error, I am unable to reproduce

>Could you fill in the description? I can merge np, but still interested to know where we're using it.

As in just the description in the PR or in the actual code?

fraction/fraction.go
```diff
@@ -0,0 +1,204 @@
...
+	gcd := f.Numerator
+	for d := f.Denominator; d != 0; {
+		gcd, d = d, gcd%d
+	}
+	f.Numerator /= gcd
+	f.Denominator /= gcd
+	return f.RectifySign()
+}
+
+// Negative - is the fractior negative
```
melekes: *fraction

fraction/fraction.go
```diff
@@ -0,0 +1,204 @@
...
+func (f fraction) Negative() bool {
+	switch {
+	case f.Numerator > 0:
+		if f.Denominator > 0 {
+			return false
+		}
+		return true
```
melekes: so if Denominator == 0, we return true? not saying this is incorrect. just checking

melekes: why not just `!f.Positive()`

fraction/fraction.go
```diff
@@ -0,0 +1,204 @@
...
+	return false
+}
+
+// Equal - test if two Fractions are equal, does not simplify
+func (f fraction) Equal(f2 Fraction) bool {
+	if f.Numerator == 0 {
+		return f2.GetNumerator() == 0
+	}
+	f1 := f.RectifySign()
+	f2 = f2.RectifySign()
```
melekes: not the best idea to mutate func args without special need...

fraction/fraction.go
```diff
@@ -0,0 +1,204 @@
...
+			f.Denominator,
+		}.Simplify()
+	}
+	return fraction{
+		f.Numerator*f2.GetDenominator() + f2.GetNumerator()*f.Denominator,
+		f.Denominator * f2.GetDenominator(),
+	}.Simplify()
+}
+
+// Sub - subtract without simplication
```
melekes: how's that `without simplication` if you call Simplify in the end?

fraction/fraction.go
```diff
@@ -0,0 +1,204 @@
...
+	if f2.GetNumerator() == 0 {
+		panic("fraction divide by zero error!")
+	}
+	return fraction{
+		f.Numerator * f2.GetDenominator(),
+		f.Denominator * f2.GetNumerator(),
+	}.Simplify()
+}
+
+// Add - add without simplication
```
melekes: how's that `without simplication` if you call Simplify in the end?

fraction/fraction.go
```diff
@@ -0,0 +1,204 @@
...
+			f.Denominator,
+		}.Simplify()
+	}
+	return fraction{
+		f.Numerator*f2.GetDenominator() - f2.GetNumerator()*f.Denominator,
+		f.Denominator * f2.GetDenominator(),
+	}.Simplify()
+}
+
+// Evaluate - evaluate the fraction using bankers rounding
```
melekes: link to the algo would be great!

fraction/fraction.go
```diff
@@ -0,0 +1,204 @@
...
+			return false
+		}
+		return true
+	}
+	return false
+}
+
+// Positive - is the fraction positive
+func (f fraction) Positive() bool {
+	switch {
```
melekes: why switch? not just `(f.Numerator > 0 && f.Denominator > 0) || (f.Numerator < 0 && f.Denominator < 0)`

jaekwon:
This package could be called "Rational".

The problem with this package is that it overflows kinda arbitrarily depending on the numbers being used as you multiply too much.  So if you multiply enough numbers together it'll just overflow.

We should just use https://golang.org/pkg/math/big/ Rat instead of this.

rigelrozanski:
@jaekwon yup! Migrating to rational usage

fraction/fraction.go
rigelrozanski: I made this algo - it's pretty simple right, round like normal, but if the last decimal place to round is the final decimal place and is a 5, then round either up or down to achieve an even number, here is a link which explains it http://www.xbeat.net/vbspeed/i_BankersRounding.htm

fraction/fraction.go
rigelrozanski: antiquated, now using big.Rat

fraction/fraction.go
rigelrozanski: antiquated, now using big.Rat

fraction/fraction.go
rigelrozanski: antiquated, now using big.Rat

fraction/fraction.go
rigelrozanski: antiquated now... interested in understanding about the reasons why it's a bad idea to mutate func arguments though!

fraction/fraction.go
rigelrozanski: because of the zero case, if zero then the result is neither neg nor positive........ Antiquated though

fraction/fraction.go
```diff
@@ -0,0 +1,204 @@
...
+
+// GetDenominator - return the Denominator
+func (f fraction) GetDenominator() int64 {
+	return f.Denominator
+}
+
+// RectifySign - make any negative sign exist only in Numerator
+func (f fraction) RectifySign() Fraction {
+	if f.Denominator < 0 {
+		return fraction{-1 * f.Numerator, -1 * f.Denominator}
```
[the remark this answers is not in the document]
rigelrozanski: antiquated

fraction/fraction.go
rigelrozanski: resolved

rational/rational.go
```diff
@@ -0,0 +1,203 @@
...
+var ten = big.NewInt(10)
+
+// EvaluateBig - evaluate the rational using bankers rounding
+func (r Rat) EvaluateBig() *big.Int {
+
+	num := r.Rat.Num()
+	denom := r.Rat.Denom()
+
+	d, rem := new(big.Int), new(big.Int)
+	d.QuoRem(num, denom, rem) // always drops the decimal
```
odeke-em: What does this comment mean? I figured that:
quotient    = num/denom  --> an int
remainder = num%denom --> an int
so perhaps this comment is not relevant?

rigelrozanski: yup comment obsolete removing. Thanks

zramsay:
closes https://github.com/cosmos/gaia/issues/9

rigelrozanski:
replaced with new PR on sdk2 branch https://github.com/tendermint/tmlibs/pull/124

PR #110, opened 2017-12-23
another timer fix

ebuchman:


common/repeat_timer.go
```diff
@@ -31,20 +33,29 @@ func (t *DefaultTicker) Chan() <-chan time.Time {
 	return t.t.C
 }
 
-// Implements Ticker
-func (t *DefaultTicker) Stop() {
+// Implements Ticker. Returns true if the ticker is stopped,
+// false if it was already stopped.
+func (t *DefaultTicker) Stop() bool {
+	if t.stopped {
```
jaekwon: race condition

jaekwon:
See https://github.com/tendermint/tmlibs/pull/111/files

common/repeat_timer.go
ebuchman: not meant to be called concurrently. 

Locking handled by the RepeatTimer

common/repeat_timer.go
```diff
@@ -8,21 +8,23 @@ import (
 // Ticker is a basic ticker interface.
 type Ticker interface {
 	Chan() <-chan time.Time
-	Stop()
+	Stop() bool
```
jaekwon: See https://github.com/tendermint/tmlibs/pull/111/files#diff-3edd8f0210da5b9853bae76c67ece2d4R203 .

FYI In tmlibs/common.Service, we return an error which is more forgiving, but seems maybe OK.  To return a bool seems to be more dangerous, as it encourages usage as if it's not an error.

PR #111, opened 2017-12-23
WIP RepeatTimer fix

jaekwon:


common/repeat_timer.go
```diff
@@ -5,152 +5,224 @@ import (
...
+		quit:   make(chan struct{}),
 	}
+	go lt.fireRoutine(interval)
+	return lt
 }
 
-// Implements Ticker
-func (t *ManualTicker) Chan() <-chan time.Time {
-	return t.ch
+// We clearly need a new goroutine, for logicalTicker may have been created
```
jaekwon: Should say, "We clearly need a new goroutine if logicalTicker.Chan() may be receiving on another routine than the source".  It's a reasonable assumption to assume that this is true, even though technically it would be possible to interleave a deterministic clock as well as waiting on logicalTicker.Chan() (with select/default) in the same go-routine, it would be a forced (unnatural) design.

Anyways, so logicalTicker starts a new go-routine that waits on the source.

This approach comes with more complexity around how it must be used.  For example, when a time gets received on the source channel, that time isn't available for anyone else, so this suggests that we should be creating a source-to-multiple-channels-multiplexer.  In other words, the implementation that is sending time to source should be doing so in a for-loop, sending the same time (or similar time) to many other receivers (e.g. more logicalTickers or sleepers).

That's fine for testing, we'll only have so many channels/goroutines as there are tendermint process goroutines that are waiting on the clock, so under a thousand, say.

common/repeat_timer.go
```diff
@@ -5,152 +5,224 @@ import (
...
 */
 type RepeatTimer struct {
-	Ch chan time.Time
+	name string
+	ch   chan time.Time
+	tm   TickerMaker
 
 	mtx    sync.Mutex
-	name   string
+	dur    time.Duration
```
jaekwon: In the future this could be updated.

common/repeat_timer_test.go
```diff
@@ -4,66 +4,86 @@ import (
...
 		default:
 		}
-		assert.True(done)
+		assert.True(t, done)
 	}
 
-	ticker := NewManualTicker(ch)
-	t := NewRepeatTimerWithTicker("bar", ticker)
+	tm := NewLogicalTickerMaker(ch)
+	dur := time.Duration(0) // dontcare
```
jaekwon: Well, we do care, but it should be under 1 second to make this test work.

common/repeat_timer_test.go
```diff
@@ -4,66 +4,86 @@ import (
...
 		for i := 0; i < cnt; i++ {
-			after := time.After(time.Second * 2)
+			timeout := time.After(time.Second * 2)
 			select {
-			case <-t.Ch:
-			case <-after:
-				test.Fatal("expected ticker to fire")
+			case _ = <-rt.Chan():
+			case <-timeout:
+				panic("QWE")
```
jaekwon: oops

common/repeat_timer_test.go
```diff
@@ -4,66 +4,86 @@ import (
...
 
-	ticker := NewManualTicker(ch)
-	t := NewRepeatTimerWithTicker("bar", ticker)
+	tm := NewLogicalTickerMaker(ch)
+	dur := time.Duration(0) // dontcare
+	rt := NewRepeatTimerWithTickerMaker("bar", dur, tm)
 
-	// start at 0
-	tock(test, t, 0)
+	// Start at 0.
```
jaekwon: This sends a time signal to the source, gets received by the logicalTicker, which sets `lasttime`.

In retrospect, what we want to test is not really the specific number of ticks, but the passage of time, to test the various time settings like commit_wait, etc.  So we maybe we need something like this:

```
func TestXYZ(t *testing.T) {
        time := logical.NewTime()
        // It's kinda nice to shadow `time` here.
        time.SetTime("2017-01-01")
        time.Start()
        defer time.Stop()

        node.SetTime(time)
        // test node
    })
}
```
And an environmental variable which could also be passed in via a flag, could set the desired time-interval for testing.  Maybe this could be overridden with time.SetClockInterval().  The only use for it I can think of is to test logical.Time itself, to initialize time.SetClockInterval() globally upon testing, or, to optimize how fast tests run, but maybe that'll never be an issue.

common/repeat_timer.go
```diff
@@ -5,152 +5,224 @@ import (
...
+		ch:     make(chan time.Time),
+		quit:   make(chan struct{}),
 	}
+	go lt.fireRoutine(interval)
+	return lt
 }
 
-// Implements Ticker
-func (t *ManualTicker) Chan() <-chan time.Time {
-	return t.ch
```
jaekwon: Since t.ch was being used as both source and `Chan()` (emitter), there was no need for another goroutine.  But the implementation is broken because a Ticker is not identical to the source.  The source is a background ticker that provides a continuous heartbeat to the program (duplicated in a for-loop for each waiting go-routine), not a ticker with a custom interval.  So we need to split this into two channels and connect them with a go-routine that filters source events before passing them on to the `.Chan()` emitter.

common/repeat_timer.go
```diff
@@ -5,152 +5,224 @@ import (
...
+	name string
+	ch   chan time.Time
+	tm   TickerMaker
 
 	mtx    sync.Mutex
-	name   string
+	dur    time.Duration
 	ticker Ticker
 	quit   chan struct{}
-	wg     *sync.WaitGroup
```
jaekwon: waitgroup not necessary with a simple modification to the fireRoutine.

jaekwon:
For future reference, here's the stack trace from Tendermint:

```
--- FAIL: TestReactor (0.56s)
panic: close of closed channel [recovered]
	panic: close of closed channel

goroutine 666 [running]:
testing.tRunner.func1(0xc4216201e0)
	/go/1.9.2/libexec/src/testing/testing.go:711 +0x2d2
panic(0x16abf40, 0x191dd40)
	/go/1.9.2/libexec/src/runtime/panic.go:491 +0x283
github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common.(*RepeatTimer).Stop(0xc42189a140, 0x0)
	/gopath/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common/repeat_timer.go:151 +0xcf
github.com/tendermint/tendermint/p2p.(*MConnection).OnStop(0xc4200fc2d0)
	/gopath/src/github.com/tendermint/tendermint/p2p/connection.go:190 +0x4b
github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common.(*BaseService).Stop(0xc4200fc2d0, 0x2, 0x2)
	/gopath/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common/service.go:145 +0x1de
github.com/tendermint/tendermint/p2p.(*peer).OnStop(0xc420456380)
	/gopath/src/github.com/tendermint/tendermint/p2p/peer.go:241 +0x33
github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common.(*BaseService).Stop(0xc420456380, 0x0, 0x0)
	/gopath/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common/service.go:145 +0x1de
github.com/tendermint/tendermint/p2p.(*Switch).OnStop(0xc420487000)
	/gopath/src/github.com/tendermint/tendermint/p2p/switch.go:219 +0x106
github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common.(*BaseService).Stop(0xc420487000, 0x17932e1, 0x2b)
	/gopath/src/github.com/tendermint/tendermint/vendor/github.com/tendermint/tmlibs/common/service.go:145 +0x1de
github.com/tendermint/tendermint/consensus.stopConsensusNet(0x1c68060, 0x1cdc668, 0xc4216406e0, 0x4, 0x4, 0xc421640720, 0x4, 0x4)
	/gopath/src/github.com/tendermint/tendermint/consensus/reactor_test.go:77 +0x172
github.com/tendermint/tendermint/consensus.TestReactor(0xc4216201e0)
	/gopath/src/github.com/tendermint/tendermint/consensus/reactor_test.go:97 +0x262
testing.tRunner(0xc4216201e0, 0x18c6b48)
	/go/1.9.2/libexec/src/testing/testing.go:746 +0xd0
created by testing.(*T).Run
	/go/1.9.2/libexec/src/testing/testing.go:789 +0x2de
```

common/repeat_timer.go
ebuchman: > logicalTicker.Chan() may be receiving on another routine than the source

Not totally sure what's meant by this.

But I see the need for the go-routine since we need to take a heartbeat and only fire when we've simulated the interval.

common/repeat_timer.go
```diff
@@ -36,7 +36,8 @@ func (t *defaultTicker) Chan() <-chan time.Time {
 
 // Implements Ticker
 func (t *defaultTicker) Stop() {
-	t.Stop()
+	tt := time.Ticker(*t)
```
jaekwon: It should be `((*time.Ticker)(t)).Stop()`. I think that one-liner is more clear.  Also I'm not sure if the above is correct.  In general we shouldn't assume that things can be dereferenced like that... often times objects are meant to be passed only by reference.  For example, see the documentation for sync.Mutex.`A Mutex must not be copied after first use.`.  In this case, nothing says a Ticker can't be copied, but it might be that the behavior is just undefined.  Whereas the one-liner is definitely going to work.

PR #113, opened 2017-12-26
Fix #112 by using RWMutex per element

### LINKED ISSUE #112: clist: rewrite code and take out WaitGroup state linked code
Inspired by undeterministic crash from this bug https://github.com/tendermint/tendermint/issues/985, the mere presence of that bug shows that our code has some underlying problems that perhaps might get triggered later on by increases in concurrency or some weird state.

I noticed that in the code we have a bunch of comments that seem to convey that allowing "races" is ok:
https://github.com/tendermint/tmlibs/blob/e3bffd8fbd9427fbfe0413ad9c813fe8b702f71f/clist/clist.go#L99

https://github.com/tendermint/tmlibs/blob/e3bffd8fbd9427fbfe0413ad9c813fe8b702f71f/clist/clist.go#L115

plus I see a bunch of logic in there where we do `*sync.WaitGroup.Add(1)` expecting `*sync.WaitGroup.Done()` and `*sync.WaitGroup.Wait()` to correspondingly be invoked. In these cases we are directly playing with racy behavior and the reported bug is a manifestation.

I'd suggest let's rewrite clist with simpler and easily reasoned about concurrency protections i.e. just mutexes.

jaekwon:
See discussion: https://github.com/tendermint/tmlibs/issues/112

### DIFF
--- a/clist/clist.go
+++ b/clist/clist.go
@@ -18,7 +18,7 @@ import (
 /*
 
 CElement is an element of a linked-list
-Traversal from a CElement are goroutine-safe.
+Traversal from a CElement is goroutine-safe.
 
 We can't avoid using WaitGroups or for-loops given the documentation
 spec without re-implementing the primitives that already exist in
@@ -220,6 +220,7 @@ func (l *CList) Front() *CElement {
 }
 
 func (l *CList) FrontWait() *CElement {
+	// Loop until the head is non-nil else wait and try again
 	for {
 		l.mtx.RLock()
 		head := l.head
@@ -230,8 +231,7 @@ func (l *CList) FrontWait() *CElement {
 			return head
 		}
 		wg.Wait()
-		// l.head doesn't necessarily exist here.
-		// That's why we need to continue a for-loop.
+		// NOTE: If you think l.head exists here, think harder.
 	}
 }
 
--- a/clist/clist.go
+++ b/clist/clist.go
@@ -11,36 +11,38 @@ to ensure garbage collection of removed elements.
 
 import (
 	"sync"
-	"sync/atomic"
-	"unsafe"
 )
 
 // CElement is an element of a linked-list
 // Traversal from a CElement are goroutine-safe.
 type CElement struct {
-	prev    unsafe.Pointer
+	mtx     sync.RWMutex
+	prev    *CElement
 	prevWg  *sync.WaitGroup
-	next    unsafe.Pointer
+	next    *CElement
 	nextWg  *sync.WaitGroup
-	removed uint32
-	Value   interface{}
+	removed bool
+
+	Value interface{} // immutable
 }
 
 // Blocking implementation of Next().
 // May return nil iff CElement was tail and got removed.
 func (e *CElement) NextWait() *CElement {
 	for {
-		e.nextWg.Wait()
-		next := e.Next()
-		if next == nil {
-			if e.Removed() {
-				return nil
-			} else {
-				continue
-			}
-		} else {
+		e.mtx.RLock()
+		next := e.next
+		nextWg := e.nextWg
+		removed := e.removed
+		e.mtx.RUnlock()
+
+		if next != nil || removed {
 			return next
 		}
+
+		nextWg.Wait()
+		// e.next doesn't necessarily exist here.
+		// That's why we need to continue a for-loop.
 	}
 }
 
@@ -48,82 +50,113 @@ func (e *CElement) NextWait() *CElement {
 // May return nil iff CElement was head and got removed.
 func (e *CElement) PrevWait() *CElement {
 	for {
-		e.prevWg.Wait()
-		prev := e.Prev()
-		if prev == nil {
-			if e.Removed() {
-				return nil
-			} else {
-				continue
-			}
-		} else {
+		e.mtx.RLock()
+		prev := e.prev
+		prevWg := e.prevWg
+		removed := e.removed
+		e.mtx.RUnlock()
+
+		if prev != nil || removed {
 			return prev
 		}
+
+		prevWg.Wait()
 	}
 }
 
 // Nonblocking, may return nil if at the end.
 func (e *CElement) Next() *CElement {
-	return (*CElement)(atomic.LoadPointer(&e.next))
+	e.mtx.RLock()
+	defer e.mtx.RUnlock()
+
+	return e.next
 }
 
 // Nonblocking, may return nil if at the end.
 func (e *CElement) Prev() *CElement {
-	return (*CElement)(atomic.LoadPointer(&e.prev))
+	e.mtx.RLock()
+	defer e.mtx.RUnlock()
+
+	return e.prev
 }
 
 func (e *CElement) Removed() bool {
-	return atomic.LoadUint32(&(e.removed)) > 0
+	e.mtx.RLock()
+	defer e.mtx.RUnlock()
+
+	return e.removed
 }
 
 func (e *CElement) DetachNext() {
 	if !e.Removed() {
 		panic("DetachNext() must be called after Remove(e)")
 	}
-	atomic.StorePointer(&e.next, nil)
+	e.mtx.Lock()
+	defer e.mtx.Unlock()
+
+	e.next = nil
 }
 
 func (e *CElement) DetachPrev() {
 	if !e.Removed() {
 		panic("DetachPrev() must be called after Remove(e)")
 	}
-	atomic.StorePointer(&e.prev, nil)
+	e.mtx.Lock()
+	defer e.mtx.Unlock()
+
+	e.prev = nil
 }
 
-func (e *CElement) setNextAtomic(next *CElement) {
-	for {
-		oldNext := atomic.LoadPointer(&e.next)
-		if !atomic.CompareAndSwapPointer(&(e.next), oldNext, unsafe.Pointer(next)) {
-			continue
-		}
-		if next == nil && oldNext != nil { // We for-loop in NextWait() so race is ok
-			e.nextWg.Add(1)
-		}
-		if next != nil && oldNext == nil {
-			e.nextWg.Done()
-		}
-		return
+// NOTE: This function needs to be safe for
+// concurrent goroutines waiting on nextWg.
+func (e *CElement) SetNext(newNext *CElement) {
+	e.mtx.Lock()
+	defer e.mtx.Unlock()
+
+	oldNext := e.next
+	e.next = newNext
+	if oldNext != nil && newNext == nil {
+		// See https://golang.org/pkg/sync/:
+		//
+		// If a WaitGroup is reused to wait for several independent sets of
+		// events, new Add calls must happen after all previous Wait calls have
+		// returned.
+		e.nextWg = waitGroup1() // WaitGroups are difficult to re-use.
+	}
+	if oldNext == nil && newNext != nil {
+		e.nextWg.Done()
 	}
 }
 
-func (e *CElement) setPrevAtomic(prev *CElement) {
-	for {
-		oldPrev := atomic.LoadPointer(&e.prev)
-		if !atomic.CompareAndSwapPointer(&(e.prev), oldPrev, unsafe.Pointer(prev)) {
-			continue
-		}
-		if prev == nil && oldPrev != nil { // We for-loop in PrevWait() so race is ok
-			e.prevWg.Add(1)
-		}
-		if prev != nil && oldPrev == nil {
-			e.prevWg.Done()
-		}
-		return
+// NOTE: This function needs to be safe for
+// concurrent goroutines waiting on prevWg
+func (e *CElement) SetPrev(newPrev *CElement) {
+	e.mtx.Lock()
+	defer e.mtx.Unlock()
+
+	oldPrev := e.prev
+	e.prev = newPrev
+	if oldPrev != nil && newPrev == nil {
+		e.prevWg = waitGroup1() // WaitGroups are difficult to re-use.
+	}
+	if oldPrev == nil && newPrev != nil {
+		e.prevWg.Done()
 	}
 }
 
-func (e *CElement) setRemovedAtomic() {
-	atomic.StoreUint32(&(e.removed), 1)
+func (e *CElement) SetRemoved() {
+	e.mtx.Lock()
+	defer e.mtx.Unlock()
+
+	e.removed = true
+
+	// This wakes up anyone waiting in either direction.
+	if e.prev == nil {
+		e.prevWg.Done()
+	}
+	if e.next == nil {
+		e.nextWg.Done()
+	}
 }
 
 //--------------------------------------------------------------------------------
@@ -132,7 +165,7 @@ func (e *CElement) setRemovedAtomic() {
 // The zero value for CList is an empty list ready to use.
 // Operations are goroutine-safe.
 type CList struct {
-	mtx  sync.Mutex
+	mtx  sync.RWMutex
 	wg   *sync.WaitGroup
 	head *CElement // first element
 	tail *CElement // last element
@@ -142,6 +175,7 @@ type CList struct {
 func (l *CList) Init() *CList {
 	l.mtx.Lock()
 	defer l.mtx.Unlock()
+
 	l.wg = waitGroup1()
 	l.head = nil
 	l.tail = nil
@@ -152,48 +186,55 @@ func (l *CList) Init() *CList {
 func New() *CList { return new(CList).Init() }
 
 func (l *CList) Len() int {
-	l.mtx.Lock()
-	defer l.mtx.Unlock()
+	l.mtx.RLock()
+	defer l.mtx.RUnlock()
+
 	return l.len
 }
 
 func (l *CList) Front() *CElement {
-	l.mtx.Lock()
-	defer l.mtx.Unlock()
+	l.mtx.RLock()
+	defer l.mtx.RUnlock()
+
 	return l.head
 }
 
 func (l *CList) FrontWait() *CElement {
 	for {
-		l.mtx.Lock()
+		l.mtx.RLock()
 		head := l.head
 		wg := l.wg
-		l.mtx.Unlock()
-		if head == nil {
-			wg.Wait()
-		} else {
+		l.mtx.RUnlock()
+
+		if head != nil {
 			return head
 		}
+		wg.Wait()
+		// l.head doesn't necessarily exist here.
+		// That's why we need to continue a for-loop.
 	}
 }
 
 func (l *CList) Back() *CElement {
-	l.mtx.Lock()
-	defer l.mtx.Unlock()
+	l.mtx.RLock()
+	defer l.mtx.RUnlock()
+
 	return l.tail
 }
 
 func (l *CList) BackWait() *CElement {
 	for {
-		l.mtx.Lock()
+		l.mtx.RLock()
 		tail := l.tail
 		wg := l.wg
-		l.mtx.Unlock()
-		if tail == nil {
-			wg.Wait()
-		} else {
+		l.mtx.RUnlock()
+
+		if tail != nil {
 			return tail
 		}
+		wg.Wait()
+		// l.tail doesn't necessarily exist here.
+		// That's why we need to continue a for-loop.
 	}
 }
 
@@ -203,11 +244,12 @@ func (l *CList) PushBack(v interface{}) *CElement {
 
 	// Construct a new element
 	e := &CElement{
-		prev:   nil,
-		prevWg: waitGroup1(),
-		next:   nil,
-		nextWg: waitGroup1(),
-		Value:  v,
+		prev:    nil,
+		prevWg:  waitGroup1(),
+		next:    nil,
+		nextWg:  waitGroup1(),
+		removed: false,
+		Value:   v,
 	}
 
 	// Release waiters on FrontWait/BackWait maybe
@@ -221,9 +263,9 @@ func (l *CList) PushBack(v interface{}) *CElement {
 		l.head = e
 		l.tail = e
 	} else {
-		l.tail.setNextAtomic(e)
-		e.setPrevAtomic(l.tail)
-		l.tail = e
+		e.SetPrev(l.tail) // We must init e first.
+		l.tail.SetNext(e) // This will make e accessible.
+		l.tail = e        // Update the list.
 	}
 
 	return e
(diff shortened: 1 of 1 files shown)


jaekwon:
@odeke-em Can you try to write a test that makes clist fail but succeed with this new PR?

odeke-em:
@jaekwon sorry, I just got back from a dinner and hangout, for sure, let me take a stab at a failing test.

clist/clist.go
```diff
@@ -1,129 +1,182 @@
...
+
+		if prev != nil || removed {
 			return prev
 		}
+
+		prevWg.Wait()
 	}
 }
 
 // Nonblocking, may return nil if at the end.
```
odeke-em: Please update this comment, this call blocks now.

clist/clist.go
```diff
@@ -1,129 +1,182 @@
...
 func (e *CElement) Next() *CElement {
-	return (*CElement)(atomic.LoadPointer(&e.next))
+	e.mtx.RLock()
+	defer e.mtx.RUnlock()
+
+	return e.next
 }
 
 // Nonblocking, may return nil if at the end.
```
odeke-em: ditto about the `non-blocking`-ness of the method.

clist/clist.go
```diff
@@ -1,129 +1,182 @@
...
+	oldNext := e.next
+	e.next = newNext
+	if oldNext != nil && newNext == nil {
+		// See https://golang.org/pkg/sync/:
+		//
+		// If a WaitGroup is reused to wait for several independent sets of
+		// events, new Add calls must happen after all previous Wait calls have
+		// returned.
+		e.nextWg = waitGroup1() // WaitGroups are difficult to re-use.
+	}
```
odeke-em: could we make this an else clause for easy readability? i.e.
```go
if oldNext != nil && newNext == nil {
   ...
} else {
   e.nextWg.Done()
}
```

clist/clist.go
```diff
@@ -1,129 +1,182 @@
...
+func (e *CElement) SetPrev(newPrev *CElement) {
+	e.mtx.Lock()
+	defer e.mtx.Unlock()
+
+	oldPrev := e.prev
+	e.prev = newPrev
+	if oldPrev != nil && newPrev == nil {
+		e.prevWg = waitGroup1() // WaitGroups are difficult to re-use.
+	}
+	if oldPrev == nil && newPrev != nil {
```
odeke-em: ditto about the else clause.

clist/clist.go
```diff
@@ -152,48 +206,55 @@ func (l *CList) Init() *CList {
...
-		if head == nil {
-			wg.Wait()
-		} else {
+		l.mtx.RUnlock()
+
+		if head != nil {
 			return head
 		}
+		wg.Wait()
+		// l.head doesn't necessarily exist here.
```
odeke-em: Perhaps we could add some more clarity to this comment?
Could we comment right before the ```for {```
```go
// Loop until the head is non-nil else wait and try again
```
and then we won't need the comment right here.

However, I believe that this is trying to reimplement the logic that sync.Cond is meant
for, moreover without the runtime polls. What do you think of us instead using sync.Cond?

clist/clist.go
```diff
@@ -1,129 +1,182 @@
...
-	"sync/atomic"
-	"unsafe"
 )
 
-// CElement is an element of a linked-list
-// Traversal from a CElement are goroutine-safe.
+/*
+
+CElement is an element of a linked-list
+Traversal from a CElement are goroutine-safe.
```
odeke-em: Minor typo/nit:
s/from a CElement are/from a CElement is/g

odeke-em:
@jaekwon please see the test below which fails both before and after, it's a case where concurrently:
* FrontWait
* Pushback
* Remove
are invoked
```go
func TestFaultyConcurrency(t *testing.T) {
        c := New()

        for i := 0; i < 1e6; i++ {
                var callCount int32
                var fini sync.WaitGroup
                fini.Add(3)
                go func() {
                        defer fini.Done()

                        atomic.AddInt32(&callCount, 1)
                        for atomic.LoadInt32(&callCount) != 3 {
                                // Spin for a bit
                        }
                        // Invokes wg.Wait()
                        c.FrontWait()
                }()

                pushbackRes := make(chan *CElement, 1)
                go func() {
                        defer fini.Done()

                        // Wait until the racy callers have been setup
                        for atomic.LoadInt32(&callCount) != 2 {
                        }
                        pushbackRes <- c.PushBack(10)
                        atomic.AddInt32(&callCount, 1)
                        // Invokes wg.Done()
                }()

                go func() {
                        defer fini.Done()

                        // Invokes wg.Add(1)
                        // but only when c.Len == 1
                        atomic.AddInt32(&callCount, 1)
                        for atomic.LoadInt32(&callCount) != 3 {
                                // Spin for a bit
                        }
                        c.Remove(<-pushbackRes)
                }()

                fini.Wait()
        }
}
```
which gives
```shell
 go test -v -run=Concurr
=== RUN   TestFaultyConcurrency
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [chan receive]:
testing.(*T).Run(0xc4200be0f0, 0x11342c6, 0x15, 0x1139df0, 0x1066501)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:825 +0x2eb
testing.runTests.func1(0xc4200be000)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:1063 +0x64
testing.tRunner(0xc4200be000, 0xc42004fdf8)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:777 +0xc4
testing.runTests(0xc42000a0c0, 0x11e13c0, 0x3, 0x3, 0x100ee89)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:1061 +0x2bb
testing.(*M).Run(0xc4200a6000, 0x0)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:978 +0x168
main.main()
	_testmain.go:46 +0x151

goroutine 5 [semacquire]:
sync.runtime_Semacquire(0xc4200ce09c)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/runtime/sema.go:56 +0x39
sync.(*WaitGroup).Wait(0xc4200ce090)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/sync/waitgroup.go:129 +0x72
github.com/tendermint/tmlibs/clist.TestFaultyConcurrency(0xc4200be0f0)
	/Users/emmanuelodeke/go/src/github.com/tendermint/tmlibs/clist/clist_test.go:265 +0x163
testing.tRunner(0xc4200be0f0, 0x1139df0)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:777 +0xc4
created by testing.(*T).Run
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:824 +0x2ca

goroutine 48 [semacquire]:
sync.runtime_Semacquire(0xc42001a11c)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/runtime/sema.go:56 +0x39
sync.(*WaitGroup).Wait(0xc42001a110)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/sync/waitgroup.go:129 +0x72
github.com/tendermint/tmlibs/clist.(*CList).FrontWait(0xc42000e3c0, 0x113a320)
	/Users/emmanuelodeke/go/src/github.com/tendermint/tmlibs/clist/clist.go:173 +0x2d
github.com/tendermint/tmlibs/clist.TestFaultyConcurrency.func1(0xc4200ce090, 0xc4200ce074, 0xc42000e3c0)
	/Users/emmanuelodeke/go/src/github.com/tendermint/tmlibs/clist/clist_test.go:238 +0x66
created by github.com/tendermint/tmlibs/clist.TestFaultyConcurrency
	/Users/emmanuelodeke/go/src/github.com/tendermint/tmlibs/clist/clist_test.go:230 +0xb7
exit status 2
FAIL	github.com/tendermint/tmlibs/clist	0.009s
Emmanuels-MacBook-Pro-2:clist emmanuelodeke$ 
Emmanuels-MacBook-Pro-2:clist emmanuelodeke$ go test -v -run=Concurr
=== RUN   TestFaultyConcurrency
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [chan receive]:
testing.(*T).Run(0xc4200be0f0, 0x1134586, 0x15, 0x113a0b0, 0x1066501)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:825 +0x2eb
testing.runTests.func1(0xc4200be000)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:1063 +0x64
testing.tRunner(0xc4200be000, 0xc42004fdf8)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:777 +0xc4
testing.runTests(0xc42000a0c0, 0x11e13c0, 0x3, 0x3, 0x100ee89)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:1061 +0x2bb
testing.(*M).Run(0xc4200a6000, 0x0)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:978 +0x168
main.main()
	_testmain.go:46 +0x151

goroutine 5 [semacquire]:
sync.runtime_Semacquire(0xc42001a29c)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/runtime/sema.go:56 +0x39
sync.(*WaitGroup).Wait(0xc42001a290)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/sync/waitgroup.go:129 +0x72
github.com/tendermint/tmlibs/clist.TestFaultyConcurrency(0xc4200be0f0)
	/Users/emmanuelodeke/go/src/github.com/tendermint/tmlibs/clist/clist_test.go:265 +0x163
testing.tRunner(0xc4200be0f0, 0x113a0b0)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:777 +0xc4
created by testing.(*T).Run
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:824 +0x2ca

goroutine 36 [semacquire]:
sync.runtime_Semacquire(0xc42001a2ac)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/runtime/sema.go:56 +0x39
sync.(*WaitGroup).Wait(0xc42001a2a0)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/sync/waitgroup.go:129 +0x72
github.com/tendermint/tmlibs/clist.(*CList).FrontWait(0xc420062740, 0x113a5e0)
	/Users/emmanuelodeke/go/src/github.com/tendermint/tmlibs/clist/clist.go:232 +0x2d
github.com/tendermint/tmlibs/clist.TestFaultyConcurrency.func1(0xc42001a290, 0xc42001a248, 0xc420062740)
	/Users/emmanuelodeke/go/src/github.com/tendermint/tmlibs/clist/clist_test.go:238 +0x66
created by github.com/tendermint/tmlibs/clist.TestFaultyConcurrency
	/Users/emmanuelodeke/go/src/github.com/tendermint/tmlibs/clist/clist_test.go:230 +0xb7
exit status 2
FAIL	github.com/tendermint/tmlibs/clist	0.011s
```

jaekwon:
Oh that's a fun one.  It's working as expected... it's just that sometimes you add an element, and then remove it, and then finally you block on FrontWait because there's nothing left anymore.

clist/clist.go
jaekwon: Not really.  All write-lock operations finish in finite expected time.  It's still non-blocking.

clist/clist.go
jaekwon: ditto dotto dotti

clist/clist.go
jaekwon: That would be wrong.

clist/clist.go
jaekwon: ditto dotto dotti

clist/clist.go
jaekwon: See https://github.com/tendermint/tmlibs/pull/113/files#diff-0326c2957f72e5e4cf728fcfdfde3521R31
I like the comment suggestion, will do.

jaekwon:
If you make those routines in a tight loop somehow I think you'll catch it though.

clist/clist.go
jaekwon:
```
!(A  &&   B)  // Same
 !A  ||  !B   // Same

so,

!((oldNext != nil)  &&  (newNext == nil)) // Same
 !(oldNext != nil)  || !(newNext == nil)  // Same
  (oldNext == nil)  ||  (newNext != nil)  // Same
```

jaekwon: Is this called Demorgan's law or is that only in the context of set unions & intersections? Anyways, same difference.

jaekwon:
BTW, `for atomic.LoadInt32(&callCount) != 2 {}` should never be written.  It's a tight loop that will waste CPU cycles doing nothing productive.  Notice that the usage of for-loops in clist.go doesn't have this problem.

What you're trying to accomplish can be done better with a second WaitGroup.

But since you probably want to write something that has three goroutines each reading, writing, and removing in a for-loop, I think you won't need that second WaitGroup.  I'm not sure how the test will look like in the end though.

clist/clist.go
odeke-em: Thanks for that catch @jaekwon, I hadn't looked at it closely. In this case, then perhaps no need for that else clause.

clist/clist.go
odeke-em: cool!

clist/clist.go
odeke-em: Thanks Jae! I understood though that head == nil at that point, I just didn't see a need for that comment hence
requested that we just make a single comment instead of the current one that has been added.

clist/clist.go
```diff
@@ -152,48 +206,55 @@ func (l *CList) Init() *CList {
...
-		if tail == nil {
-			wg.Wait()
-		} else {
+		l.mtx.RUnlock()
+
+		if tail != nil {
 			return tail
 		}
+		wg.Wait()
+		// l.tail doesn't necessarily exist here.
```
odeke-em: I don't think that this comment is necessary, which is why like in the previous I requested we just add a comment to the top say
```go
// Loop until the tail is non-nil else try again.
```

odeke-em:
@jaekwon 

 BTW, for atomic.LoadInt32(&callCount) != 2 {} should never be written. It's a tight loop that will 
 waste CPU cycles doing nothing productive. Notice that the usage of for-loops in clist.go doesn't 
 have this problem.

that `atomic.LoadInt32(&callCount) != 2` is meant to ensure that both the goroutines(before and after) are synchronized and have reached the point for us to trigger the concurrent actions otherwise we won't be able to reach the point. What's the reasoning behind it never being written? Could you perhaps show me an alternate way of synchronizing the two other goroutines to the desired point?

jaekwon:
It should never be written because we should never write logic that is not guaranteed to terminate eventually.  If you add a `runtime.GoSched()` call in that tight loop it would be more correct.  Please read up on what GoSched() is and when it's needed.

But even still, it's not a good idea to write such tight loops.  While `GoSched()` helps, it still doesn't guarantee much... I'm not sure that it even guarantees non-chokingness, and even if it does, it's still not a very strong guarantee.

You're still burning cycles doing nothing.  If this were in production code in any of Google's major long running services, it would cost them a fortune.  In practice it will cause unexpected indeterminism, and if this code snippet were to find its way into production non-test code, we'd see a *major* performance degradation.  It's therefore a bad pattern and should never be used, even in testing.

You want goroutine #2 to "wait" until #1 and #3 are "done" starting up (new WaitGroup "A"), and you want #1 and #3 to "wait" until #2 is "done" sending on `pushbackRes` (new WaitGroup "B")
 (btw, please append `Ch` to channels in general, so much preferred is `pushbackResCh`).

So the first problem is that you're spinning without `GoSched()`, but more importantly that you're spinning at all.  The second problem is that what you're doing with for-loop under `// Wait until the racy callers have been setup` is completely pointless.  In other words, WaitGroup "A" is unnecessary and contributes nothing.  This is because after both goroutine #1 and #3 are "done" starting up, there is no guarantee about when the next line of code will execute, so it's exactly the same as if nothing was done at all.  Of course both goroutine #1 and #3 added a defer statement, but the call to `Done()` in `defer fini.Done()` doesn't actually get called until later, so the behavior and guarantees are exactly the same as if WaitGroup "A" weren't there.  In otherwords, you might as well do:

```golang
               go func() {
                        defer fini.Done()

                        for atomic.LoadInt32(&callCount) != 1 {
                                // Spin for a bit
                        }
                        // Invokes wg.Wait()
                        c.FrontWait()
                }()

                pushbackRes := make(chan *CElement, 1)
                go func() {
                        defer fini.Done()

                        pushbackRes <- c.PushBack(10)
                        atomic.AddInt32(&callCount, 1)
                        // Invokes wg.Done()
                }()

                go func() {
                        defer fini.Done()

                        for atomic.LoadInt32(&callCount) != 1 {
                                // Spin for a bit
                        }
                        c.Remove(<-pushbackRes)
                }()
```

jaekwon:
If the defer statement looked like the following, it would be a different story: `defer fini.Something().Done()`, because then `Something()` would get called immediately, and then the `Done()` would be deferred.

Now if you change the remaining spinning logic to use a WaitGroup, then I think it'll become clearer what the third and final problem is.

jaekwon:
This should be required reading for everyone:

https://golang.org/ref/mem

odeke-em:
@jaekwon will all due respect I disagree in a few aspects. For sure there could be more precision with the ideas you put up such as using runtime.Gosched(), but I don't buy that Gosched() is needed at all for this simple test, and I think you might be complicating the problem/solution.

Firstly, for a straw man, that test is adapted directly from the Go source tree which actually runs at Google multiple times a day and a bunch of their code has such tests https://github.com/golang/go/blob/acce8268b611613bef4c422a861bd4863e85b0f1/src/sync/waitgroup_test.go#L71-L125. 

For sure spinning burns CPU massively, but the point here is a simple test that will recreate the race condition and it is guaranteed to terminate at least from the structure of it above, unless am mistaken -- logically there is no way that callCount will never be 3 unless only one goroutine is ran and the others are deliberately stopped which I believe would require hacking the runtime.
 
The fini waitGroup is created once per loop and it is there to ensure that ALL that goroutines have returned before we proceed to the next iteration. I don't have more than one waitgroup at all in a single loop, which waitgroups are you talking about?

However, I feel like I am being defensive and perhaps missing the point rather than learning and writing correct code. Please feel free to suggest a more efficient/proper test.

jaekwon:
> Firstly, for a straw man, that test is adapted directly from the Go source tree which actually runs at Google multiple times a day and a bunch of their code has such tests https://github.com/golang/go/blob/acce8268b611613bef4c422a861bd4863e85b0f1/src/sync/waitgroup_test.go#L71-L125.

Nay, the only strawman here is your example.  In the next test they do exactly what I suggested (`TestWaitGroupMisuse3`), and the reason why they use a wasteful spin-lock for `TestWaitGroupMisuse2` is because that test works even when the implementation of WaitGroup is broken.  So literally the only reason to use wasteful spin-locks is to test your implementation of locking primitives.  If you have examples of wasteful spin-locks outside the context of testing locking primitives, that would be interesting.

> For sure spinning burns CPU massively, but the point here is a simple test that will recreate the race condition and it is guaranteed to terminate at least from the structure of it above, unless am mistaken -- logically there is no way that callCount will never be 3 unless only one goroutine is ran and the others are deliberately stopped which I believe would require hacking the runtime.

The point is that there's clearly a better and more efficient way to do it.  For any programmer, `TestWaitGroupMisuse3` is MUCH easier to understand than `TestWaitGroupMisuse2`, and if you disagree, I will happily put my money where my mouth is and make a bet with you w/ very favorable odds for you.  I've also written other reasons why it's bad, and those reasons still hold.  

> I don't have more than one waitgroup at all in a single loop, which waitgroups are you talking about?

I'm saying that the correct way to write your test is to use 3 WaitGroups instead of 1 WaitGroup and 3 wasteful spin-locks.  Then I showed you why actually instead of 3 you only need 2.  I leave it as a challenge to you to figure out what that looks like.  All the information is written up there.

jaekwon:
> a bunch of their code has such tests 

You'll have to point to them because I can't find any at first glance.

https://github.com/golang/go/search?utf8=%E2%9C%93&q=%2F%2F+spin&type=

jaekwon:
Here's a clearer explanation of what I wrote.

https://groups.google.com/forum/#!msg/golang-nuts/eu9h0bT5hGI/xzB0hi1Y1tYJ

```golang
var a string
var done bool

func setup() {
	a = "hello, world"
	done = true
}

func main() {
	go setup()
	for !done {
	}
	print(a)
}
```

`print(a)` is not guaranteed to run.  The https://golang.org/ref/mem link explains why.

Also, in the example you wrote, all three goroutines spin after an `atomic.Add` call, so it's correct, but your code is *almost* broken.  `pushbackRes <- c.PushBack(10)` is NOT guaranteed to run before `c.FrontWait()`.  (Can you see why?)  It so happens that it doesn't matter, but you can see how it's very tricky to get right.

odeke-em:
>...
print(a) is not guaranteed to run. The https://golang.org/ref/mem link explains why.

```go
go setup()
for !done {
}
print(a)
```

there is an explicit atomic data check, not a plain `for !done` like you are likening it to be, the runtime could schedule the goroutine setup() to run later and by that time `for !done` could have terminated, that's a well known problem. However, `for atomic.Load*` while being a tight loop, ensures that we spin until we actually observe a write, it's a synchronization mechanism. I mean a seemingly better and guaranteed way would be using a channel but I don't think that suffices here.

>...
pushbackRes <- c.PushBack(10) is NOT guaranteed to run before c.FrontWait(). (Can you see why?)

Using the buffered channel is NOT guaranteed to run before c.FrontWait(), but the point of this test is to probabilistically clash `c.Remove(<-pushbackRes)` with `c.FrontWait()` and I used the buffered channel to ensure that as soon as
```go
atomic.AddInt32(&callCount, 1)
```
executed, a race would eventually be triggered, waiting for its reader to pull it ASAP and invoke `c.Remove`.

Anyways, I think my mental and other budgets are spent on the back and forth, it's taking more time to go back and forth on this than actually fixing the underlying problems.

PR #114, opened 2017-12-26
WIP Changed Service interface

jaekwon:
Service could be much simplified.
We shouldn't be masking race conditions by letting the caller call Start/Stop willy nilly.  If one needs to rely on such patterns, it's probably because one has racy buggy software.

Now that we know better, lets refactor and test Tendermint to see how racy it is.

We don't use Restart as far as I know.

### DIFF
--- a/common/service.go
+++ b/common/service.go
@@ -6,39 +6,30 @@ import (
 	"github.com/tendermint/tmlibs/log"
 )
 
-type Service interface {
-	Start() (bool, error)
+type ServiceCore interface {
 	OnStart() error
+	OnStop() error
+}
 
-	Stop() bool
-	OnStop()
-
-	Reset() (bool, error)
-	OnReset() error
-
+type Service interface {
+	Start() error
+	Stop() error
 	IsRunning() bool
-
 	String() string
-
+	SetServiceCore(ServiceCore)
 	SetLogger(log.Logger)
 }
 
 /*
-Classical-inheritance-style service declarations. Services can be started, then
-stopped, then optionally restarted.
 
-Users can override the OnStart/OnStop methods. In the absence of errors, these
-methods are guaranteed to be called at most once. If OnStart returns an error,
-service won't be marked as started, so the user can call Start again.
+Services can be started, then stopped, then optionally restarted.
 
-A call to Reset will panic, unless OnReset is overwritten, allowing
-OnStart/OnStop to be called again.
+Users can override the OnStart/OnStop methods.
+These methods are guaranteed to be called at most once.
+Starting an already started service will panic.
+Stopping an already stopped (or non-started) service will panic.
 
-The caller must ensure that Start and Stop are not called concurrently.
-
-It is ok to call Stop without calling Start first.
-
-Typical usage:
+Usage:
 
 	type FooService struct {
 		BaseService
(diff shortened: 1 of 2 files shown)


common/service.go
```diff
@@ -6,39 +6,30 @@ import (
...
-
-Users can override the OnStart/OnStop methods. In the absence of errors, these
-methods are guaranteed to be called at most once. If OnStart returns an error,
-service won't be marked as started, so the user can call Start again.
-
-A call to Reset will panic, unless OnReset is overwritten, allowing
-OnStart/OnStop to be called again.
 
-The caller must ensure that Start and Stop are not called concurrently.
+Services can be started, then stopped, then optionally restarted.
```
jaekwon: Oops, no it can't be restarted.

common/service.go
```diff
@@ -6,101 +6,90 @@ import (
...
+// SetServiceCore allows a user to set a ServiceCore. Users should implement their logic using
+// ServiceCore.
+type Service interface {
+	Start() error
+	Stop() error
 	IsRunning() bool
-
-	String() string
-
+	SetServiceCore(ServiceCore)
```
odeke-em: I like the new usage of `ServiceCore`. However, shouldn't we also make ServiceCore implement the String() method if Service implicitly invokes OnStart() and OnStop() of the ServiceCore?

common/service.go
```diff
@@ -110,79 +99,44 @@ func (bs *BaseService) Start() (bool, error) {
...
+// IsRunning implements Service
 func (bs *BaseService) IsRunning() bool {
 	return atomic.LoadUint32(&bs.started) == 1 && atomic.LoadUint32(&bs.stopped) == 0
 }
 
-func (bs *BaseService) Wait() {
-	<-bs.Quit
+// SetServiceCore impleents SetServiceCore
+func (bs *BaseService) SetServiceCore(service ServiceCore) {
+	bs.impl = service
```
odeke-em: While I see an atomic protection for started and stopped, there is none for logger and impl which sure by contract will only be set once, but if we've got mutual exclusion for some fields, why not for all of them?
This makes sure that our service our bullet proof and not reliant on the user, who can potentially breach the contract.

adrianbrink:
@odeke-em I like the idea of protecting everything. What do you think about just removing SetLogger and SetServiceCore from the interface. Both need to be passed into the constructor and then it can only be started and stopped.

jaekwon:
Just noting that we'll want to expose the quit chan somehow.  Maybe `Quit() <- struct{}`

common/service.go
ebuchman: The String is just the name passed into NewBaseService

common/service.go
ebuchman: We should clarify which things should and shouldn't be called multiple times or concurrently in the comments, but it might be confusing to make things thread safe when they shouldnt be

ebuchman (commented):
I like the simplification. PR still needs some work.

ebuchman:
@greg-szabo let's add this to next release (not 0.16) and discuss on monday

ebuchman:
damn we forgot to talk about this. maybe we do need to use an organization level kanban @greg-szabo @jolesbi

jolesbi:
@greg-szabo can you put it on the tendermint meeting agenda for next week? alternatively, someone can set up a call to discuss. #zoomzoom 🚗

ebuchman:
We should move the Service to Tendermint. It's the only consumer AFAIK.

Same with the timers!

zramsay:
this PR is now captured in https://github.com/tendermint/tendermint/issues/1882 but let's leave it open for now so the branch doesn't get deleted, or close and delete if obsolete

PR #119, opened 2017-12-30
Sdk2

ebuchman:
Major improvements to `db`, mostly around iterators


common/bytes.go
```diff
@@ -0,0 +1,53 @@
...
+	return nil
+}
+
+// This is the point of Bytes.
+func (bz HexBytes) MarshalJSON() ([]byte, error) {
+	s := strings.ToUpper(hex.EncodeToString(bz))
+	jbz := make([]byte, len(s)+2)
+	jbz[0] = '"'
+	copy(jbz[1:], []byte(s))
+	jbz[1] = '"'
```
ebuchman: uh

db/fsdb.go
```diff
@@ -0,0 +1,254 @@
...
+	panic("FSDB.Stats not yet implemented")
+}
+
+func (db *FSDB) NewBatch() Batch {
+	db.mtx.Lock()
+	defer db.mtx.Unlock()
+
+	// Not sure we would ever want to try...
+	// It doesn't seem easy for general filesystems.
+	panic("FSDB.NewBatch not yet implemented")
```
ebuchman: reminder for go-crypto keybase.Update. need to Set new before deleting old

PR #121, opened 2018-01-13
WIP: common/random.go supports seeding and *Rand

jaekwon:


jaekwon:
Test was broken, fixed the test to do the opposite of what it was trying -- namely, ensure that output is deterministic.

jaekwon:
From the previous tests:  

> // Set math/rand's Seed so that any direct invocations of math/rand will reveal themselves.

How does setting the seed help reveal any invocations of math/rand?

It would make sense if the function first called math.rand.Seed(1), then *consumed* math/rand.Int*() to measure what it would produce, so that it can then reset to math.rand.Seed(1) *again*, then called cmn.Rand*(), then consumed math/rand.Int*() *again*, to ensure that the output from the first consumption still equaled the output from the second consumption.

But the determinism test does the same thing, if you assume that the consumption of math/rand.* gets returned rather than being thrown away.

odeke-em:
>How does setting the seed help reveal any invocations of math/rand?

Once one sets rand.Seed(1), if there are any internal usages of rand.*, they'll produce predictable outputs so for example if we invoke multiple times, our internal PRNG's RandStr invokes math/rand.RandInt63(); math/rand.Seed ensures that such usages produce deterministic outputs -- basically that any places that use math/rand.* will have a common seed and our tests can then catch them.

>It would make sense if the function first called math.rand.Seed(1), then consumed math/rand.Int*() to measure what it would produce, so that it can then reset to math.rand.Seed(1) again, then called cmn.Rand*(), then consumed math/rand.Int*() again, to ensure that the output from the first consumption still equaled the output from the second consumption.

Sure, that's actually a great and direct way to test for it. However, the previous test sort of did that because it reset itself in https://github.com/tendermint/tmlibs/pull/121/files#diff-c4914c202a8cbd7b81b27ef73a1d30aaL58 and invoked math/rand.Seed in https://github.com/tendermint/tmlibs/pull/121/files#diff-c4914c202a8cbd7b81b27ef73a1d30aaL62 and if any function used math/rand, we'd get a clash thereafter. I like your new suggestion better.

By the counting PRNG I just realized that that isn't even possible with math/rand, am bullshitting there. If we were using crypto/rand https://play.golang.org/p/wVuXl5pLb-g would be relevant.

Anyways, I see what your new test does and we can incorporate your new suggestion to compare output from math/rand and math/rand.Seed since my attack scenario would only work for crypto/rand.Reader as above

melekes:
What's left here? Can we merge?

PR #123, opened 2018-01-14
Merge sdk2 into develop

ebuchman:


cli/setup.go
```diff
@@ -42,7 +39,7 @@ func PrepareBaseCmd(cmd *cobra.Command, envPrefix, defaultHome string) Executor
 func PrepareMainCmd(cmd *cobra.Command, envPrefix, defaultHome string) Executor {
 	cmd.PersistentFlags().StringP(EncodingFlag, "e", "hex", "Binary encoding (hex|b64|btc)")
```
melekes: should we delete this flag too if we're removing `setEncoding`?

PR #124, opened 2018-01-18
Rational

rigelrozanski:
- [x] fix go-wire marshalling 
 - [x] add rational interface


rational/rational.go
```diff
@@ -0,0 +1,203 @@
...
+//TODO there has got to be a better way using native MarshalText and UnmarshalText
+
+// RatMarshal - Marshable Rat Struct
+type RatMarshal struct {
+	Numerator, Denominator int64
+}
+
+// MarshalJSON - custom implementation of JSON Marshal
+func (r Rat) MarshalJSON() ([]byte, error) {
+	return json.Marshal(RatMarshal{r.Num(), r.Denom()})
```
odeke-em:
```
wire.MarshalJSON(RatMarshal{r.Num(), r.Denom()})
```

rational/rational.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+// UnmarshalJSON - custom implementation of JSON Unmarshal
+func (r *Rat) UnmarshalJSON(data []byte) error {
+	ratMar := new(RatMarshal)
+	err := json.Unmarshal(data, ratMar)
```
odeke-em:
```
if err := wire.UnmarshalJSON(data, ratMar); err != nil {
   return err
}
```

rigelrozanski:
@odeke-em `wire.marshalJSON` works fine for a struct with exposed fields, what I was hoping to accomplish was overriding the default marshal procedure in go-wire with custom marshal functions `MarshalJSON` and `UnmarshalJSON` so I would still be able to use go-wire for structs and interfaces which include a `rational` field within them. There is some weird dep issues so I isolated my rational branch to this repo temporarily: https://github.com/rigelrozanski/rational-test

melekes:
Is this still WIP?

rigelrozanski:
@melekes You're right no longer WIP as it should be working... I guess let's merge soon? My original thinking was to leave open and merge simultaneously with the WIP PR of Staking Module into `cosmos-sdk` but maybe it actually makes more sense to merge this in now, it's done unless others have further critique or Rational.. but we can always update with more PR - so yes I'd say let's merge. However further considerations:
This branch depends on some new `sdk2` branch stuff in go-wire, it should probably be rebased and updated As soon as @jaekwon merges Go-Wire sdk2 branch into `develop` then this branch should be rebased to develop I guess... So we just have to consider the merge order, but I'd like to get this in sooner than later

rational/rational.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	str := strings.Split(decimalStr, ".")
+
+	var numStr string
+	var denom int64 = 1
+	switch len(str) {
+	case 1:
+		if len(str[0]) == 0 {
+			return f, fmt.Errorf("not a decimal string")
```
melekes: could've just used `errors.New`

rational/rational.go
```diff
@@ -0,0 +1,221 @@
...
+	return r.EvaluateBig().Int64()
+}
+
+// Round - round Rat with the provided precisionFactor
+func (r Rat) Round(precisionFactor int64) Rational {
+	rTen := Rat{new(big.Rat).Mul(r.Rat, big.NewRat(precisionFactor, 1))}
+	return Rat{big.NewRat(rTen.Evaluate(), precisionFactor)}
+}
+
+//// Old Reference code
```
melekes: do we need this?

rational/rational.go
```diff
@@ -0,0 +1,221 @@
...
+// MarshalJSON - custom implementation of JSON Marshal
+func (r Rat) MarshalJSON() ([]byte, error) {
+	return cdc.MarshalJSON(RatMarshal{r.Num(), r.Denom()})
+}
+
+// UnmarshalJSON - custom implementation of JSON Unmarshal
+func (r *Rat) UnmarshalJSON(data []byte) (err error) {
+	defer func() {
+		if recover() != nil {
+			err = errors.New("divide by zero error during Unmarshal")
```
melekes: how do you know it's divide by zero error?

rational/rational.go
rigelrozanski: okay using `errors.New` I usually just use fmt.Errors to avoid confusing with other error packages that come up

rational/rational.go
rigelrozanski: nah deleted

melekes:
```
--- FAIL: TestZeroSerializationJSON (0.00s)
        Error Trace:    rational_test.go:212
        Error:          Expected nil, but got: &errors.errorString{s:"Unexpected json bytes, expected an opaque hex-string as a stub."}
        Test:           TestZeroSerializationJSON
--- FAIL: TestSerializationGoWire (0.00s)
        Error Trace:    rational_test.go:248
        Error:          Should be true
        Test:           TestSerializationGoWire
        Messages:       original: 1/3, unmarshalled: 0/1
--- FAIL: TestEmbeddedStructSerializationGoWire (0.00s)
        Error Trace:    rational_test.go:271
        Error:          Should be true
        Test:           TestEmbeddedStructSerializationGoWire
        Messages:       original: {foo 10 1/3}, unmarshalled: {foo 10 0/1}
--- FAIL: TestEmbeddedInterfaceSerializationGoWire (0.00s)
        Error Trace:    rational_test.go:295
        Error:          Should be true
        Test:           TestEmbeddedInterfaceSerializationGoWire
        Messages:       original: {foo 10 1/3}, unmarshalled: {foo 10 0/1}
FAIL
```

does it require one of the PRs here https://github.com/tendermint/go-wire/pulls ?

rational/rational.go
rigelrozanski: updated in to include the recover message, I didn't think it would panic during anything else, either way updated now

rigelrozanski:
@melekes yes need new go-wire stuff here: https://github.com/tendermint/go-wire/pull/83

jolesbi:
hey @rigelrozanski @melekes! can we fix CI here so when https://github.com/tendermint/go-wire/pull/83 gets merged we're ready to rock'n'roll? 🎸🥁

rigelrozanski:
@jolesbi yeah I think the only reason CI broken is because of a funky inter-dep issue - I'll take a peak but I'm pretty sure CI is blocked on the earlier merge

rigelrozanski:
@melekes werks now - but we should probably still wait for go-wire updates

ebuchman:
Ahh this didn't hit me before. But we just finished removing dependency on go-wire from tmlibs! It would be wonderful to not bring it back. 

So maybe rational just belongs in the SDK ?

ebuchman:
Or it can stay here if we can remove go-wire from it

rigelrozanski:
@ebuchman I think it needs go-wire so we can marshal's which embeds a rational type... Maybe into the types directory of the SDK?

rigelrozanski:
pulled into https://github.com/cosmos/cosmos-sdk/pull/492

PR #128, opened 2018-01-22
Hashable -> Hasher; SimpleMap upgrade; No "SimpleHashFromBinary"

jaekwon:
Update SimpleMap to hash both keys and values for benefit:
* You can blind the key even with a Merkle proof.
* It's potentially shorter than a long key
* Always hash the value, drop "SimpleHashFromBinary(ies)"
Hashable is Hasher
Don't assume go-wire


merkle/simple_map.go
```diff
@@ -18,25 +18,25 @@ func NewSimpleMap() *SimpleMap {
...
 
-	// Is value Hashable?
-	var vBytes []byte
-	if hashable, ok := value.(Hashable); ok {
-		vBytes = hashable.Hash()
-	} else {
-		vBytes, _ = wire.MarshalBinary(value)
-	}
+	// Hash the key to blind it... why not?
+	khash := SimpleHashFromBytes([]byte(key))
```
melekes: why not declare `key` as `Hasher` too and call `key.Hash()` here?

merkle/simple_tree.go
```diff
@@ -54,28 +54,25 @@ func SimpleHashFromHashes(hashes [][]byte) []byte {
 	}
 }
 
-// Convenience for SimpleHashFromHashes.
-func SimpleHashFromBinaries(items []interface{}) []byte {
-	hashes := make([][]byte, len(items))
-	for i, item := range items {
-		hashes[i] = SimpleHashFromBinary(item)
+// NOTE: Do not implement this, use SimpleHashFromByteslices instead.
```
melekes: Do not implement what?

merkle/types.go
```diff
@@ -18,6 +18,10 @@ type Tree interface {
 	IterateRange(start []byte, end []byte, ascending bool, fx func(key []byte, value []byte) (stop bool)) (stopped bool)
 }
 
-type Hashable interface {
+type Hasher interface {
 	Hash() []byte
 }
+
+type Byteser interface {
```
melekes: Do we need to define this interface?

merkle/simple_map.go
```diff
@@ -18,25 +18,25 @@ func NewSimpleMap() *SimpleMap {
...
-		Value: vBytes,
+		Key:   khash,
+		Value: vhash,
 	})
 }
 
-// Merkle root hash of items sorted by key.
-// NOTE: Behavior is undefined when key is duplicate.
+// Merkle root hash of items sorted by key
+// (UNSTABLE: and by value too if duplicate key).
```
ebuchman: UNSTABLE ?

merkle/simple_map.go
jaekwon: I think most cases will just need a string or []byte, so optimizing for the typical case.

merkle/simple_map.go
jaekwon: behavior may change.

merkle/simple_tree.go
jaekwon: SimpleHashFromBytesers.  Better to just use SimpleHashFromByteslices.

merkle/types.go
jaekwon: No.

PR #132, opened 2018-01-25
v0.7.0 release

zramsay:
small (but breaking) fixe for tendermint config


melekes:
If this is a new code, it should go into `develop` first

PR #136, opened 2018-01-28
common: fix BitArray.Update to avoid nil dereference

odeke-em:
Update previously only checked that the receiver was
non-nil but didn't check that the input parameter to update
"o" was non-nil causing a nil dereference in cases such as
https://github.com/tendermint/tendermint/blob/fe632ea32a89c3d9804bbd6e3ce9391b1d5a0993/consensus/reactor.go#L306

Fixes https://github.com/tendermint/tendermint/issues/1169

### DIFF
--- a/common/bit_array_test.go
+++ b/common/bit_array_test.go
@@ -164,3 +164,26 @@ func TestEmptyFull(t *testing.T) {
 		}
 	}
 }
+
+func TestUpdateNeverPanics(t *testing.T) {
+	newRandBitArray := func(n int) *BitArray {
+		ba, _ := randBitArray(n)
+		return ba
+	}
+	pairs := []struct {
+		a, b *BitArray
+	}{
+		{nil, nil},
+		{newRandBitArray(10), newRandBitArray(12)},
+		{newRandBitArray(0), NewBitArray(10)},
+		{nil, NewBitArray(10)},
+		{nil, newRandBitArray(64)},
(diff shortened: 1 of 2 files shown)


common/bit_array_test.go
```diff
@@ -164,3 +164,26 @@ func TestEmptyFull(t *testing.T) {
...
+	pairs := []struct {
+		a, b *BitArray
+	}{
+		{nil, nil},
+		{newRandBitArray(10), newRandBitArray(12)},
+		{newRandBitArray(0), NewBitArray(10)},
+		{nil, NewBitArray(10)},
+		{nil, newRandBitArray(64)},
+		{newRandBitArray(63), newRandBitArray(64)},
+	}
```
adrianbrink: Could you simplify the test cases:
* nil, nil,
* newRandBitArray, newRandBitArray
* newRandBitArray, nil
* nil, newRandBitArray

That way we test all the possible combination while also making it easier for developers to read and understand this test.

odeke-em: Great idea, thanks @adrianbrink, I'll update that right now.

odeke-em: Done, with https://github.com/tendermint/tmlibs/pull/136/commits/85be26c675b05a2a75c856f7c22b446d8df1c944

odeke-em:
Thank you for the review @adrianbrink!

PR #142, opened 2018-02-03
common: IsHex should be able to handle 0X prefixed strings

odeke-em:
IsHex should also successfully decode strings prefixed with
0X instead of only 0x strings.

Also add tests generally for IsHex.

### DIFF
--- a/common/string.go
+++ b/common/string.go
@@ -29,7 +29,7 @@ func LeftPadString(s string, totalLength int) string {
 
 // IsHex returns true for non-empty hex-string prefixed with "0x"
 func IsHex(s string) bool {
-	if len(s) > 2 && s[:2] == "0x" {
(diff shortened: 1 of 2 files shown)


common/string.go
```diff
@@ -29,7 +29,7 @@ func LeftPadString(s string, totalLength int) string {
 
 // IsHex returns true for non-empty hex-string prefixed with "0x"
 func IsHex(s string) bool {
-	if len(s) > 2 && s[:2] == "0x" {
+	if len(s) > 2 && strings.EqualFold(s[:2], "0x") {
```
rigelrozanski: I guess this is to allow for "0x" OR "OX"

odeke-em:
Thanks @melekes and @rigelrozanski for the review and merge!

Issue #146, opened 2018-02-06
common/BitArray: Or is buggy with a nil deference and mistaken nil check clause instead of a return


odeke-em:
I was just auditing the code and noticed that *BitArray.Or's code looks like this
https://github.com/tendermint/tmlibs/blob/19e818fdf79aa9290661acfe45a3f4dfd4e996a9/common/bit_array.go#L102-L104

Notice the bug there
if bA == nil, we invoke o.Copy(), of which o might also be nil. But the biggest problem even is that even if bA, we intended to return o.Copy() but instead forgot about the return and on the next line invoke bA.mtx.Lock() which delivers a crash to us.

We really need to make rigorous tests as it seems that we've been testing with the best conditions intended but in the wild we find the actual edge cases e.g nil arguments being passed in as in https://github.com/tendermint/tendermint/issues/1169

odeke-em:
Fixed by #147 which has been merged into develop.

Issue #145, opened 2018-02-06
common/BitArray: all setters whose arguments take in another *BitArray should first check for nil-ity


odeke-em:
I've been auditing usages of BitArray and found that similiar to this embarrassing bug https://github.com/tendermint/tendermint/issues/1169 that we fixed with https://github.com/tendermint/tmlibs/pull/136 we should check to see that the other *BitArray being passed in is non-nil

A simple exhibit is this test that fails at head https://github.com/tendermint/tmlibs/commit/19e818fdf79aa9290661acfe45a3f4dfd4e996a9 after
```diff
diff --git a/common/bit_array_test.go b/common/bit_array_test.go
index e4ac8bf..3d4b3b3 100644
--- a/common/bit_array_test.go
+++ b/common/bit_array_test.go
@@ -46,6 +46,8 @@ func TestOr(t *testing.T) {
 	bA2, _ := randBitArray(31)
 	bA3 := bA1.Or(bA2)
 
+	bA1.Or(nil)
+
 	if bA3.Bits != 51 {
 		t.Error("Expected max bits")
 	}
```

with stack trace
```shell
$ go test -v -run=Or
=== RUN   TestOr
--- FAIL: TestOr (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x12a71cb]

goroutine 20 [running]:
testing.tRunner.func1(0xc4201560f0)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:742 +0x29d
panic(0x1304e40, 0x152a4f0)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/runtime/panic.go:505 +0x229
github.com/tendermint/tmlibs/common.(*BitArray).Or(0xc4200a3620, 0x0, 0x0)
	/Users/emmanuelodeke/go/src/github.com/tendermint/tmlibs/common/bit_array.go:107 +0x7b
github.com/tendermint/tmlibs/common.TestOr(0xc4201560f0)
	/Users/emmanuelodeke/go/src/github.com/tendermint/tmlibs/common/bit_array_test.go:49 +0x8c
testing.tRunner(0xc4201560f0, 0x13818d0)
	/Users/emmanuelodeke/go/src/go.googlesource.com/go/src/testing/testing.go:777 +0xd0
[...]

PR #147, opened 2018-02-06
common/BitArray: reduce fragility with methods

odeke-em:
Fixes https://github.com/tendermint/tmlibs/issues/145
Fixes https://github.com/tendermint/tmlibs/issues/146

The code in here has been fragile when it comes to nil
but these edge cases were never tested, although they've
showed up in the wild and were only noticed because
the reporter actually read the logs otherwise
we'd have never known.

This changes covers some of these cases and adds some tests.


odeke-em:
Thanks @melekes and @ebuchman for the review and merge!

PR #156, opened 2018-03-02
pubsub implements service.OnReset

simonvadee:
This is needed for tendermint/tendermint#1239

### DIFF
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -209,6 +209,11 @@ func (s *Server) OnStart() error {
 	return nil
 }
 
+// OnReset implements Service.OnReset
+func (s *Server) OnReset() error {
+	return nil
+}
+
 func (s *Server) loop(state state) {
 loop:
 	for cmd := range s.cmds {


pubsub/pubsub.go
```diff
@@ -200,12 +200,22 @@ type state struct {
...
+
+// OnStart implements Service.OnStart by starting the server.
+func (s *Server) OnStart() error {
+	s.startLoop()
+	return nil
+}
+
+// OnReset implements Service.OnReset by starting the server.
+func (s *Server) OnReset() error {
+	s.startLoop()
```
melekes: Reset should not Start anything. It should be empty.

So normal sequence:

```
Stop()
Reset()
Start()
```

simonvadee:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[pubsub/pubsub.go, line 218 at r1](https://reviewable.io:443/reviews/tendermint/tmlibs/156#-L6aBighrobY3H-jAeqi-r1-218:-L6anknrsie-1hBhzIec:b-296dqg) ([raw file](https://github.com/tendermint/tmlibs/blob/69947afb3774d0043a7ce89be1e20d0d25556ada/pubsub/pubsub.go#L218)):*
<details><summary><i>Previously, melekes (Anton Kaliaev) wrote…</i></summary><blockquote>

Reset should not Start anything. It should be empty.

So normal sequence:

```
Stop()
Reset()
Start()
```
</blockquote></details>

you're right, my bad ! fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/tendermint/tmlibs/156)*

PR #158, opened 2018-03-08
WIP - CLI refactor to work with REST

### LINKED ISSUE #157: Make CLI work better for REST initialization
All the CLI stuff is good for setting up a CLI - REST functionality is almost identical however we don't want to use cobra, only viper config.

rigelrozanski:
closes #157


ebuchman:
Is anything using this lib besides the SDK? 

If not, let's move it into the SDK. I don't see good reason for it to be here ... it's too magical and has been a source of confusion and bugs before because it's so remote ...

rigelrozanski:
@ebuchman It's a part of tendemint core...

melekes:
> it's too magical and has been a source of confusion and bugs before because it's so remote ...

agree, less magic the better

faboweb:
If I could wish for sth. it would be tests. :-)

Wish

ebuchman:
What exactly is this doing for us ?

rigelrozanski:
I don't even remember, it was something I made when working with Fabo on REST stuff - I think this no longer being used... @faboweb speak now or forever hold yer peace - I'll close if I don't hear back soon

Issue #159, opened 2018-03-11
common: WriteFileAtomic should not ignore file.Write errors while trying to preserve error before cleanup


odeke-em:
The code for WriteFileAtomic attempts to always cleanup on any error but unfortunately it has a flaw in that it ignores the error from f.Write

https://github.com/tendermint/tmlibs/blob/1b9b5652a199ab0be2e781393fb275b66377309d/common/os.go#L135

We can avoid such by:
a) Making a named error `err` 
b) A defer cleanup routine that'll clean up the created directory if err != nil 

The above will properly handle our conditions and let's also try to return early on errors.

odeke-em:
I discovered while investigating https://github.com/tendermint/tendermint/issues/1290. We are close to launch hence should be more careful with such code plus anything that Tendermint uses.

odeke-em:
Ironically for WriteFileAtomic, if f.Write returns an error, no other error check will be performed because of the error checking style
as per 
https://github.com/tendermint/tmlibs/blob/1b9b5652a199ab0be2e781393fb275b66377309d/common/os.go#L135-L153

melekes:
> it has a flaw in that it ignores the error from f.Write

no, it does not.

odeke-em:
That was poor wording from me, thanks for pointing that out @melekes. It takes the error from f.Write if that occurs but doesn't return immediately thus ignores all the other errors, say if Close , permission errors occur.

odeke-em:
@xla addressed this with https://github.com/tendermint/tmlibs/pull/171.

PR #160, opened 2018-03-11
common: WriteFileAtomic should not ignore Write error; make named return

odeke-em:
Fixes https://github.com/tendermint/tmlibs/issues/159

On encountering an error after file.Write, instead of returning
all along we've been proceeding with the code below it, invoking
* file.Close()
* os.Chmod
while ignoring their errors ironically, because *Atomic implies
"all or nothing" but that's contrary to the current behavior.

This change fixes that with a named return and returning early.


common/os.go
```diff
@@ -126,30 +126,42 @@ func MustWriteFile(filePath string, contents []byte, mode os.FileMode) {
...
+
+	defer func() {
+		// Any error should result in full cleanup.
+		// See Issue https://github.com/tendermint/tmlibs/issues/159
+		if err != nil {
+			// If the file happens to be open, let's close it
+			// lest be hung up on it such as on Windows
+			// where we need to first close before removing
+			// See Issue https://github.com/golang/go/issues/16296.
+			if f != nil {
```
melekes: nice find! why just don't add this to 

```
// any err should result in full cleanup
if err != nil {
  <HERE>
  os.Remove(f.Name())	
}
```

melekes:
What's the benefit of returning early? a few nanoseconds? plus I bet the old version is actually faster because now we have a `defer`.

melekes:
Plus if we're gonna change this code which is battle tested, we should add more tests `62.48% (-0.21%) compared to 066fe82`

odeke-em:
>What's the benefit of returning early? a few nanoseconds? plus I bet the old version is actually faster because now we have a defer.

Not speed related at all, it is so that we can "ATOMICALLY"(all or nothing) check for any failures e.g if f.Write fails, we should exit. If f.Close() fails we should exit, if the permission change fails, return early.

odeke-em:
Ping @melekes @ebuchman.

melekes:
> "ATOMICALLY"(all or nothing)

this is true for the current version

common/os.go
odeke-em: The goal here is to close if the file is non-nil before os.Remove. I should have instead moved `os.Remove` in the `if f != nil` check

xla:
Can we close this in favour of #171 ?

ebuchman:
Thanks @odeke-em - let's go with #171

PR #161, opened 2018-03-11
Fix byteslice encoding for SimpleMap

jaekwon:
When we removed the go-wire dependency, ByteSlice was using varints.  We now use uvarints exclusively to encode the length of items to follow.

We probably won't have these issues anymore, but it's worth noting that we almost forgot to fix this.... we'll just need to write more integration tests.

### DIFF
--- a/events/README.md
+++ b/events/README.md
@@ -95,7 +95,7 @@ type EventCallback func(data EventData)
 type EventData interface {
 }
 ```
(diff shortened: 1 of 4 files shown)


ebuchman:
:o nice catch! Thanks.

I copied these functions over to remove the dependency on go-wire. Dangerous!

PR #162, opened 2018-03-12
DB as a service: remove database deployment with API

odeke-em:
Fixes https://github.com/tendermint/tendermint/issues/1162

Databases as a service!

Can now access Databases as a remote service
via gRPC for performance and easy deployment.

The caveat is that each service is stateful in regards to
the DB i.e. each unique service uses only one unique DB
but nonetheless multiple clients can access it.

A full standalone/end-to-end example in a test

```go
package remotedb_test

import (
	"net"
	"testing"

	"github.com/stretchr/testify/require"

	"github.com/tendermint/tmlibs/grpcdb"
	"github.com/tendermint/tmlibs/remotedb"
)

func TestRemoteDB(t *testing.T) {
	ln, err := net.Listen("tcp", "0.0.0.0:0")
	require.Nil(t, err, "expecting a port to have been assigned on which we can listen")
	srv := grpcdb.NewServer()
	defer srv.Stop()
	go func() {
		if err := srv.Serve(ln); err != nil {
			t.Fatalf("BindServer: %v", err)
		}
	}()

	client, err := remotedb.NewInsecure(ln.Addr().String())
	require.Nil(t, err, "expecting a successful client creation")
	require.Nil(t, client.InitRemote(&remotedb.Init{Name: "test-remote-db", Type: "leveldb"}))

	k1 := []byte("key-1")
	v1 := client.Get(k1)
	require.Equal(t, 0, len(v1), "expecting no key1 to have been stored")
	vv1 := []byte("value-1")
	client.Set(k1, vv1)
	gv1 := client.Get(k1)
	require.Equal(t, gv1, vv1)

	// Deletion
	client.Delete(k1)
	gv2 := client.Get(k1)
	require.Equal(t, len(gv2), 0, "after deletion, not expecting the key to exist anymore")
	require.NotEqual(t, len(gv1), len(gv2), "after deletion, not expecting the key to exist anymore")
}
```


grpcdb/server.go
```diff
@@ -0,0 +1,162 @@
...
+func (s *server) Iterator(query *protodb.Entity, dis protodb.DB_IteratorServer) error {
+	it := s.db.Iterator(query.Start, query.End)
+	return s.handleIterator(it, dis.Send)
+}
+
+func (s *server) handleIterator(it db.Iterator, sendFunc func(*protodb.Iterator) error) error {
+	for it.Valid() {
+		start, end := it.Domain()
+		out := &protodb.Iterator{
+			Domain: &protodb.DDomain{Start: start, End: end},
```
melekes: `DDomain` as Database Domain?

proto/defs.proto
```diff
@@ -0,0 +1,57 @@
...
+message Entity {
+  int32 id	= 1;
+  bytes key	= 2;
+  bytes value	= 3;
+  bool exists	= 4;
+  bytes start	= 5;
+  bytes end	= 6;
+  string err	= 7;
+  string print	= 8;
+  int64 time_at = 9;
```
melekes: does `time_at` mean `created_at` or `received_at` or ?

proto/defs.proto
```diff
@@ -0,0 +1,57 @@
...
+
+message Entity {
+  int32 id	= 1;
+  bytes key	= 2;
+  bytes value	= 3;
+  bool exists	= 4;
+  bytes start	= 5;
+  bytes end	= 6;
+  string err	= 7;
+  string print	= 8;
```
melekes: seems to be unused

melekes:
Looks 🆒

proto/defs.proto
cwgoes: `created_at`, updated.

proto/defs.proto
cwgoes: Removed.

grpcdb/server.go
cwgoes: Changed to just `Domain`.

cwgoes:
(new issue - ref https://github.com/tendermint/tmlibs/issues/203)

grpcdb/client.go
```diff
@@ -0,0 +1,30 @@
...
+type Security uint
+
+const (
+	Insecure Security = iota
+	Secure
+)
+
+// NewClient creates a gRPC client connected to the bound gRPC server at serverAddr.
+// Use kind to set the level of security to either Secure or Insecure.
+func NewClient(serverAddr string, serverCert string) (protodb.DBClient, error) {
```
melekes: `serverAddr, serverCert string`

grpcdb/doc.go
```diff
@@ -0,0 +1,32 @@
...
+	defer srv.Stop()
+	go func() {
+		if err := srv.Serve(ln); err != nil {
+			t.Fatalf("BindServer: %v", err)
+		}
+	}()
+
+or
+	addr := ":8998"
+  cert := "server.crt"
```
melekes: something wrong with formatting

grpcdb/server.go
```diff
@@ -0,0 +1,171 @@
...
+	"google.golang.org/grpc/credentials"
+
+	"github.com/tendermint/tmlibs/db"
+	protodb "github.com/tendermint/tmlibs/proto"
+)
+
+// ListenAndServe is a blocking function that sets up a gRPC based
+// server at the address supplied, with the gRPC options passed in.
+// Normally in usage, invoke it in a goroutine like you would for http.ListenAndServe.
+func ListenAndServe(addr string, cert string, key string, opts ...grpc.ServerOption) error {
```
melekes: same here `addr, cert, key string`

grpcdb/server.go
```diff
@@ -0,0 +1,171 @@
...
+		return err
+	}
+	srv, err := NewServer(cert, key, opts...)
+	if err != nil {
+		return err
+	}
+	return srv.Serve(ln)
+}
+
+func NewServer(cert string, key string, opts ...grpc.ServerOption) (*grpc.Server, error) {
```
melekes: same

grpcdb/server.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (s *server) Delete(ctx context.Context, in *protodb.Entity) (*protodb.Nothing, error) {
+	s.db.Delete(in.Key)
+	return nothing, nil
+}
+
+var nothing = new(protodb.Nothing)
+
+func (s *server) DeleteSync(ctx context.Context, in *protodb.Entity) (*protodb.Nothing, error) {
```
melekes: why return `nothing`?

grpcdb/client.go
cwgoes: Addressed.

grpcdb/server.go
cwgoes: Addressed.

grpcdb/server.go
cwgoes: Addressed.

grpcdb/server.go
cwgoes: gRPC needs a value, maybe?

grpcdb/doc.go
cwgoes: Fixed.

cwgoes:
Batch operations implemented, addressed comments. We might want more tests still.

remotedb/remotedb.go
```diff
@@ -218,5 +220,43 @@ func (itr *iterator) Value() []byte {
 }
 
 func (itr *iterator) Close() {
-	// TODO: Shut down the iterator
+	err := itr.dic.CloseSend()
```
melekes: weird method name... does `CloseSend` mean close for sending?

cwgoes: Close the send-side, I think. Modeled after the example at the bottom of https://grpc.io/docs/tutorials/basic/go.html.

odeke-em:
Great, thank you @cwgoes for shipping this and @melekes for the review!

grpcdb/server.go
odeke-em: Yes, the protobuf compiler can't generate an rpc definition that without a return value i.e.
```go
service Foo {
   rpc Bar() 
}
```  
but it'll accept
```go
service Foo {
   rpc Foo(Nothing) returns (Nothing)
}
```

PR #163, opened 2018-03-12
Option to disable time prefixing

odeke-em:
First part of https://github.com/tendermint/tendermint/issues/1096

A non-time prefixed logger can be created by passing in
an option
```go
lg := log.NewTMLogger(bufNoTimePrefix, log.OptionDisableTimePrefix)
```
while without that option, you'll have time prefixing
```go
lg := log.NewTMLogger(bufNoTimePrefix)
```

and added tests too.

### DIFF
--- a/log/tm_logger.go
+++ b/log/tm_logger.go
@@ -24,7 +24,7 @@ var _ Logger = (*tmLogger)(nil)
 // NewTMTermLogger returns a logger that encodes msg and keyvals to the Writer
 // using go-kit's log as an underlying logger and our custom formatter. Note
 // that underlying logger could be swapped with something else.
-func NewTMLogger(w io.Writer) Logger {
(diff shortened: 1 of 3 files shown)


melekes:
This is good, but I think we can improve it by abstracting the format https://github.com/tendermint/tendermint/issues/1096#issuecomment-357326560

xla:
This change in itself might not be all that useful as discussed in tendermint/tendermint#1096 and tendermint/tendermint#1212 should be part of a more directed effort to streamline logging.

Two other observations, it feels exceptional to have a single flag controlling a single field, instead of having a general way of determining what fields are part of a log line. And the way timestamps are prefixed breaks with the gains of structured logs, smells a bit like the 90s.

90s

odeke-em:
Roger that, thank you both for the input, I'll close this PR and defer to the consensus in those issues.

PR #164, opened 2018-03-14
Implement NewPrefixDB

jaekwon:
To fix the MultiStore.


db/mem_batch.go
```diff
@@ -35,16 +37,29 @@ func (mBatch *memBatch) Delete(key []byte) {
...
 	defer mtx.Unlock()
 
-	for _, op := range mBatch.ops {
-		switch op.opType {
-		case opTypeSet:
-			mBatch.db.SetNoLock(op.key, op.value)
-		case opTypeDelete:
-			mBatch.db.DeleteNoLock(op.key)
+	for i, op := range mBatch.ops {
+		if i < len(mBatch.ops) {
```
melekes: I don't understand how's i can be greater than `len(mBatch.ops)` if we're ranging over it

jaekwon: oops, you're right, I missed a -1 . I'll add a test.

PR #165, opened 2018-03-14
Fix flowrate test

jaekwon:
NOTE: PR is against jae/dbprefix, not develop.

I discovered this bug while running the test...


jaekwon:
Related: https://github.com/mxk/go-flowrate/issues/3

PR #168, opened 2018-03-18
common: remove {Left, Right}PadString

### LINKED ISSUE #134: common: please remove LeftPad* and RightPad* as they are unused, or at least replace them
These helpers seem unused throughout the entire organization and are somewhat expensive/less concise.

https://github.com/tendermint/tmlibs/blob/91b4b534ad78e442192c8175db92a06a51064064/common/string.go#L12-L28

can be converted into

```go
func RightPadString(s string, totalLength uint) string {
       return fmt.Sprintf("% *s", totalLength, s)
}

func LeftPadString(s string, totalLength uint) string {
       return fmt.Sprintf("% -*s", totalLength, s)
}
```

odeke-em:
Fixes #134

Those functions are unused in the whole Tendermint Github
organization plus they were unnecessariy verbose and could
have been concisely replaced with

```go
func RightPadString(s string, totalLength uint) string {
       return fmt.Sprintf("% *s", totalLength, s)
}

func LeftPadString(s string, totalLength uint) string {
       return fmt.Sprintf("% -*s", totalLength, s)
}
```

delete them anyways


melekes:
🥇 🏎 🍰

PR #170, opened 2018-03-18
common: NewBitArray never crashes on negatives

### LINKED ISSUE #169: common: NewBitArray should check for negatives and not just 0
Currently the code  in NewBitArray is trusting and assumes that rational and trusting users will use this code. However, Byzantine actors don't care, they'll send us negative lengths if they can and in deed, for example in https://github.com/tendermint/tendermint/issues/1322 they can.

https://github.com/tendermint/tmlibs/blob/1b9b5652a199ab0be2e781393fb275b66377309d/common/bit_array.go#L16-L25

If they pass in lengths <=-127, game over for us. Usually the verbal defence would be "just audit all the spots that use this code and ensure they don't pass in negatives" but I'll argue that this code could be made more secure just by changing:
`if bits == 0`
to
`if bits <= 0`

The following test should never crash
```go
func TestBitArrayCatchesNegative(t *testing.T) {
        bitList := []int{-127, -128}
        for _, bits := range bitList {
                _ = common.NewBitArray(bits)
        }
}
```

odeke-em:
Fixes #169
Fixes https://github.com/tendermint/tendermint/issues/1322

The previous code was very trusting assuming that
rational actors will use this code. However, Byzantine
actors don't care and in the case of the linked issue
negative lengths can be sent to this code unfettered
having been received from a peer.

This code is essentially just a sign change from
`==`
to
`<=`

and we've gutted out that attack by being more defensive.


melekes:
👍 nice find

odeke-em:
Thank you for the review and merge @melekes!

PR #171, opened 2018-03-18
Simplify WriteFileAtomic

xla:
We can make the implementation more robust by adjusting our assumptions and leverage explicit file modes for syncing. Additionally we going to assume that we want to clean up and can't really recover if thos operations (file close and removal) fail.

* utilise file mode for majority of concerns
* improve test coverage by covering more assumptions
* signature parity with ioutil.WriteFile
* always clean up

Replaces #160

### DIFF
--- a/common/os.go
+++ b/common/os.go
@@ -124,32 +124,32 @@ func MustWriteFile(filePath string, contents []byte, mode os.FileMode) {
 	}
 }
 
-// WriteFileAtomic writes newBytes to temp and atomically moves to filePath
-// when everything else succeeds.
-func WriteFileAtomic(filePath string, newBytes []byte, mode os.FileMode) error {
-	dir := filepath.Dir(filePath)
-	f, err := ioutil.TempFile(dir, "")
+// WriteFileAtomic creates a temporary file with data and the perm given and
+// swaps it atomically with filename if successful.
+func WriteFileAtomic(filename string, data []byte, perm os.FileMode) error {
+	var (
+		dir      = filepath.Dir(filename)
+		tempFile = filepath.Join(dir, "write-file-atomic-"+RandStr(32))
+		// Override in case it does exist, create in case it doesn't and force kernel
+		// flush, which still leaves the potential of lingering disk cache.
+		flag = os.O_WRONLY | os.O_CREATE | os.O_SYNC | os.O_TRUNC
+	)
+
+	f, err := os.OpenFile(tempFile, flag, perm)
 	if err != nil {
 		return err
 	}
-	_, err = f.Write(newBytes)
-	if err == nil {
-		err = f.Sync()
+	// Clean up in any case. Defer stacking order is last-in-first-out.
+	defer os.Remove(f.Name())
+	defer f.Close()
+
+	if n, err := f.Write(data); err != nil {
+		return err
+	} else if n < len(data) {
+		return io.ErrShortWrite
 	}
-	if closeErr := f.Close(); err == nil {
-		err = closeErr
-	}
-	if permErr := os.Chmod(f.Name(), mode); err == nil {
-		err = permErr
-	}
-	if err == nil {
-		err = os.Rename(f.Name(), filePath)
-	}
-	// any err should result in full cleanup
-	if err != nil {
-		os.Remove(f.Name())
-	}
-	return err
+
+	return os.Rename(f.Name(), filename)
 }
 
 //--------------------------------------------------------------------------------
--- a/common/os_test.go
+++ b/common/os_test.go
@@ -2,29 +2,51 @@ package common
 
 import (
 	"bytes"
-	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"testing"
 	"time"
 )
 
 func TestWriteFileAtomic(t *testing.T) {
-	data := []byte("Becatron")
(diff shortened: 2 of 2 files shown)


common/os.go
```diff
@@ -124,32 +124,32 @@ func MustWriteFile(filePath string, contents []byte, mode os.FileMode) {
...
-// when everything else succeeds.
-func WriteFileAtomic(filePath string, newBytes []byte, mode os.FileMode) error {
-	dir := filepath.Dir(filePath)
-	f, err := ioutil.TempFile(dir, "")
+// WriteFileAtomic creates a temporary file with data and the perm given and
+// swaps it atomically with filename if successful.
+func WriteFileAtomic(filename string, data []byte, perm os.FileMode) error {
+	var (
+		dir      = filepath.Dir(filename)
+		tempFile = filepath.Join(dir, "write-file-atomi-"+RandStr(32))
```
melekes: `atomic`

common/os.go
```diff
@@ -124,32 +124,32 @@ func MustWriteFile(filePath string, contents []byte, mode os.FileMode) {
...
+		flag = os.O_WRONLY | os.O_CREATE | os.O_SYNC | os.O_TRUNC
+	)
+
+	f, err := os.OpenFile(tempFile, flag, perm)
 	if err != nil {
 		return err
 	}
-	_, err = f.Write(newBytes)
-	if err == nil {
-		err = f.Sync()
```
melekes: when will syncing take place in the new implementation? after `os.Rename` during `f.Close()`?

xla: Write will not return until `sync` is complete, thanks to `O_SYNC` that we pass to `OpenFile`:

https://github.com/tendermint/tmlibs/pull/171/files#diff-989bc07942973b256a849bf6b6b2788fR135

common/os.go
xla: Thanks, good catch.

common/os.go
```diff
@@ -124,32 +124,32 @@ func MustWriteFile(filePath string, contents []byte, mode os.FileMode) {
...
-		err = permErr
-	}
-	if err == nil {
-		err = os.Rename(f.Name(), filePath)
-	}
-	// any err should result in full cleanup
-	if err != nil {
-		os.Remove(f.Name())
+	// Clean up in any case.
+	defer f.Close()
```
odeke-em: This order isn't going to work on Windows, it'll leave the file undeleted. You want to first close then remove.

common/os.go
```diff
@@ -124,32 +124,32 @@ func MustWriteFile(filePath string, contents []byte, mode os.FileMode) {
...
-// when everything else succeeds.
-func WriteFileAtomic(filePath string, newBytes []byte, mode os.FileMode) error {
-	dir := filepath.Dir(filePath)
-	f, err := ioutil.TempFile(dir, "")
+// WriteFileAtomic creates a temporary file with data and the perm given and
+// swaps it atomically with filename if successful.
+func WriteFileAtomic(filename string, data []byte, perm os.FileMode) error {
+	var (
+		dir      = filepath.Dir(filename)
+		tempFile = filepath.Join(dir, "write-file-atomic-"+RandStr(32))
```
odeke-em: Am a little hesitant on this new approach, the previous `ioutil.TempFile` handled that already with all the permissions and wouldn't add the tempfile to the same working directory as this.

xla: The old implementation used the directory of the target file, this is why issues like tendermint/tendermint#1329 surface. Sticking to the current approach might just be the best as we need to expect proper rights to be able to write to that directory, any other target directory might be too much of an assumption on the environment.

common/os.go
xla: Great refresher on defer stack order, thanks! Will address.

common/os.go
odeke-em: >The old implementation used the directory of the target file, this is why issues like 

Thank you for the correction @xla, for some reason I assumed that since ioutil.TempDir adds them to the tempdir, vicariously ioutil.Tempfile did too, but I should have known better ;) However, your approach adds the file to be renamed to the final destination's directory too with `filepath.Join(parentDir, generatedSuffix)`, right?
I am just asking to understand the change better.

common/os.go
```diff
@@ -124,32 +124,32 @@ func MustWriteFile(filePath string, contents []byte, mode os.FileMode) {
...
-	if permErr := os.Chmod(f.Name(), mode); err == nil {
-		err = permErr
-	}
-	if err == nil {
-		err = os.Rename(f.Name(), filePath)
-	}
-	// any err should result in full cleanup
-	if err != nil {
-		os.Remove(f.Name())
+	// Clean up in any case. Defer stacking order is last-in-first-out.
```
odeke-em: Cool. For a minor nit, in the future, we never know who might not get the order of these so maybe
```go
defer func() {
    f.Close()
    os.Remove(f.Name())
}()
```

common/os.go
xla: `ioutil.TempDir` only uses `os.Tempdir` if the dir argument is left blank, same for `ioutil.TempFile`.

> However, your approach adds the file to be renamed to the final destination's directory too with filepath.Join(parentDir, generatedSuffix), right?

Indeed, while it is tempting to resort to the platform specific temp directories, it lacks guarantees, from the docs:

> The directory is neither guaranteed to exist nor have accessible permissions.

common/os.go
xla: Could be slightly improved, I believe for now the comment suffices.

PR #174, opened 2018-03-20
Implement better Parallel

jaekwon:
In the few places where we use Parallel, we probably also want a way to abort the call (which dispatches the multiple go-routines).

For example, here: https://github.com/tendermint/tendermint/blob/develop/p2p/conn/secret_connection.go#L205
we never actually return upon err1 because the reader (the second task, aka #1) would block forever (or until connection timeout, but that's not what we really want).

```go
	var err1, err2 error
	cmn.Parallel(
		func() {
			_, err1 = conn.Write(locEphPub[:])
		},
		func() {
			remEphPub = new([32]byte)
			_, err2 = io.ReadFull(conn, remEphPub[:])
		},
	)
	if err1 != nil {
		return nil, err1
	}
	if err2 != nil {
		return nil, err2
	}
```

The new Parallel() lets each task optionally return abort=true, which lets the parent call return asap.


common/async.go
```diff
@@ -1,15 +1,59 @@
...
+	// respective taskResultCh (associated by task index).
+	for i, task := range tasks {
+		var taskResultCh = make(chan TaskResult, 1) // Capacity for 1 result.
+		taskResultChz[i] = taskResultCh
+		go func(i int, task Task, taskResultCh chan TaskResult) {
+			// Recovery
+			go func() {
+				if pnk := recover(); pnk != nil {
+					taskResultCh <- TaskResult{nil, nil, pnk}
+				}
```
melekes: so, if we panic, we never write true to `taskDoneCh` and Parallel never finishes?

xla:
Have we looked into alternative, already maintained methods to handle parallel execution, like https://github.com/oklog/run?

common/async.go
jaekwon: oh we should write falase to taskDoneCh.

jaekwon:
oklog/run is not general enough.

```go
// Run all actors (functions) concurrently.
// When the first actor returns, all others are interrupted.
// Run only returns when all actors have exited.
// Run returns the error returned by the first exiting actor.
```

Has different semantics than this Parallel.  This is a general unit that can be combined with other units (like an interruptor or quit channel), but it has the benefit that it doesn't prescribe where it doesn't need to.  For example, Parallel here does not return only when all tasks have completed.  It returns (or at least provides a way to receive) all errors, not just the first error, etc.

I haven't seen any good async.Parallel implementations in Golang, I did do a brief scan.  In any case I'm pretty confident that this is a unique design, and I'm partial to this design (or in general to any design that is likewise general without prescribing any more than necessary).  Happy to compare with other libraries, but we'd have to talk about specifics of design, not just whether it supports parallel execution and is maintained/active.

xla:
@jaekwon Thanks for the thorough insight. I see now how our requirements are distinct from most async implementations.

common/async.go
jaekwon: that's strange, I wonder why the panic test didn't catch that.  I'll take a look.

jaekwon: looks like the tests did catch it, I just never ran them properly :P
Please checkout the latest commit.

PR #180, opened 2018-03-24
New Error

jaekwon:
This replaces pkg/errors.
Intent is so we can call cmn.ErrorWrap instead of errors.Wrap.
This should be more performant, as it doesn't require capturing a stacktrace each wrap.


common/errors.go
```diff
@@ -5,16 +5,54 @@ import (
...
+	Message() string
+	Stacktrace() Error
 	Trace(format string, a ...interface{}) Error
-	TraceCause(cause error, format string, a ...interface{}) Error
-	Cause() error
-	Type() interface{}
+	TraceAt(n int, format string, a ...interface{}) Error
+	Cause() interface{}
 	WithType(t interface{}) Error
+	Type() interface{}
```
jaekwon: Is there a better name?  Switcher?  
`switch err.Type().(type)` is redundant...
`switch err.Switcher().(type)` is also redundant...

common/errors.go
```diff
@@ -5,16 +5,54 @@ import (
 	"runtime"
 )
 
+//----------------------------------------
+// Convenience methods
+
+// ErrorWrap will just call .TraceAt(), or create a new *cmnError.
+func ErrorWrap(cause interface{}, format string, a ...interface{}) Error {
```
melekes: `args ...` please

common/errors.go
```diff
@@ -5,16 +5,54 @@ import (
...
+		// NOTE: cause may be nil.
+		// NOTE: do not use causeCmnError here, not the same as nil.
+		return newError(msg, cause, cause).Stacktrace()
+	}
+}
+
 //----------------------------------------
 // Error & cmnError
 
+/*
```
melekes: why not move this to `errors` subpackage? then it can be `errors.Wrap`

common/errors.go
```diff
@@ -5,16 +5,54 @@ import (
...
+*/
 type Error interface {
 	Error() string
+	Message() string
+	Stacktrace() Error
 	Trace(format string, a ...interface{}) Error
-	TraceCause(cause error, format string, a ...interface{}) Error
-	Cause() error
-	Type() interface{}
+	TraceAt(n int, format string, a ...interface{}) Error
```
melekes: `TraceFrom(level int` or `TraceFrom(depth int`

common/errors.go
melekes: I'd vote for the first option (maybe even shorten it to `err.T().(type)`. go is redundant language (`obj.(Object).SomeMethod`).

common/errors.go
```diff
@@ -24,119 +62,169 @@ func NewError(format string, a ...interface{}) Error {
...
 }
 
 type cmnError struct {
-	msg    string
-	cause  error
-	type_  interface{}
-	traces []traceItem
+	msg        string         // first msg which also appears in msg
+	cause      interface{}    // underlying cause (or panic object)
+	type_      interface{}    // for switching on error
```
melekes: what's up with `_`?

common/errors.go
```diff
@@ -24,119 +62,169 @@ func NewError(format string, a ...interface{}) Error {
...
+
 func (err *cmnError) Error() string {
-	return fmt.Sprintf("Error{%v:%s,%v,%v}", err.type_, err.msg, err.cause, len(err.traces))
+	return fmt.Sprintf("%v", err)
+}
+
+// Captures a stacktrace if one was not already captured.
+func (err *cmnError) Stacktrace() Error {
+	if err.stacktrace == nil {
+		err.stacktrace = captureStacktrace(3, 32)
```
melekes: magic numbers

common/errors.go
jaekwon: can't use "type" as field name

common/errors.go
jaekwon: I like "T" a lot.  It differentiates it from the "type" keyword.

common/errors.go
jaekwon: k

common/errors.go
jaekwon: tmlibs errors?  I'm down for that.

common/errors.go
jaekwon: sure

common/errors.go
jaekwon: the 3 I think arguably should be magic, though it deserves comments.  I'll move the 32 out.

melekes: you don't have to move them out, you can do

```
var offset = 3
const defaultDepth = 32
```

inside the function

common/errors.go
jaekwon: well, because cmn is convenient...  if we want to have cmn, then it's more convenient to do cmn.ErrorWrap, than importing another module.

jaekwon: I'd argue that Go2 should support Error natively... so I'd want it in cmn.

melekes: Yeah.. `cmn.ErrorWrap` is fine too

PR #182, opened 2018-03-28
Return configuration parse errors instead of printing them to stderr

cwgoes:
Most clients of this library probably want to handle errors themselves and inform the user that their config is invalid.

Ref: https://github.com/cosmos/cosmos-sdk/issues/717


cwgoes:
AFAIK this can be merged.

PR #189, opened 2018-04-03
bug fix: WriteFileAtomic

mohanson:
Must close file before rename it.

### DIFF
--- a/common/os.go
+++ b/common/os.go
@@ -149,8 +149,7 @@ func WriteFileAtomic(filename string, data []byte, perm os.FileMode) error {
 		return io.ErrShortWrite
 	}
 	// Close the file before renaming it, otherwise it will cause "The process 
-	// cannot access the file because it is being used by another process." on windows or 
-	// cause "cross-link error" on linux when you try to save it to another partition.
+	// cannot access the file because it is being used by another process." on windows.
 	f.Close()
 
 	return os.Rename(f.Name(), filename)
--- a/common/os.go
+++ b/common/os.go
@@ -148,6 +148,9 @@ func WriteFileAtomic(filename string, data []byte, perm os.FileMode) error {
 	} else if n < len(data) {
 		return io.ErrShortWrite
 	}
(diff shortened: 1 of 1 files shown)


melekes:
> bug fix

what bug does this fix?

> Must close file before rename it.

how did you figure this out? can you come up with a test?

common/os.go
```diff
@@ -148,6 +148,7 @@ func WriteFileAtomic(filename string, data []byte, perm os.FileMode) error {
 	} else if n < len(data) {
 		return io.ErrShortWrite
 	}
+	f.Close()
```
melekes: can you add a comment here explaining why we need to do that.

like "close the file before renaming it. otherwise ...."

mohanson: In the same device with the same partition type, the "rename" on linux only modify the file name, so
it will not affect the existing read and write operations. But If you "rename" it to another device, it will crash ("cross-link error").
This problem may also exist in another operating systems.

common/os.go
```diff
@@ -148,6 +148,9 @@ func WriteFileAtomic(filename string, data []byte, perm os.FileMode) error {
 	} else if n < len(data) {
 		return io.ErrShortWrite
 	}
+	// Close the file before renaming it, otherwise it will cause "The process 
+	// cannot access the file because it is being used by another process." on windows or 
+	// cause "cross-link error" on linux when you try to save it to another partition.
```
melekes: note we're using the same dir where filename is located, so cross-link error can't happen

mohanson: yes.

ebuchman:
Notice we have a `defer f.Close()`, so in the absence of error we'll try to close the file twice.

I'm not sure that's actually a problem, but worth pointing out - and if we ever check the errors on Close(), I think the second one will be non-nil

ebuchman:
https://github.com/tendermint/tmlibs/issues/192

faboweb:
Is this on tendermint#develop already?

ebuchman:
it is now

PR #190, opened 2018-04-03
Fix lint errors

tomc974:
Same as I did for the `tendermint` repo, this PR is just fixing a bunch of lint errors. Each error category has it's own commit so if any one of them doesn't suit your style, it will be easy to revert it.


melekes:
@tomc974 thank you for working on this!

PR #195, opened 2018-04-05
Use an interface for tags.

### LINKED ISSUE #105: pubsub: use an interface with Get() and Len() for tags
rather than a map

tomc974:
Fixes #105

The issue was a little vague, so I'm not sure the fix is quite  what was expected.
Also  I could not come with a better name than `ITagSet` which is not quite the Go way.


melekes:
@tomc974 thank you for working on this!

pubsub/pubsub.go
```diff
@@ -38,18 +38,25 @@ var (
 	ErrAlreadySubscribed = errors.New("already subscribed")
 )
 
+type ITagSet interface {
```
melekes: how about

```
// TagMap ....
type TagMap interface {
        // Get returns the value for a key, or nil if no value is present. The ok result indicates whether value was found in the tags.
        Get(key string) (value interface{}, ok bool)
        // Len returns the number of tags
	Len() int
}

type tagMap map[string]interface{}

....
```

pubsub/pubsub.go
```diff
@@ -38,18 +38,25 @@ var (
...
+	Get(key string) (result interface{}, found bool)
+	Len() int
+}
+
+type TagSet map[string]interface{}
```
melekes: do we need to export this?? see my comment above

pubsub/pubsub.go
tomc974: Will do that!

pubsub/pubsub.go
tomc974: Don't know what it is used for but if you think it should be private so be it.

pubsub/pubsub.go
```diff
@@ -38,18 +38,28 @@ var (
 	ErrAlreadySubscribed = errors.New("already subscribed")
 )
 
+type TagMap interface {
```
melekes: could you please add a comment

tomc974: done!

melekes:
Wanna upgrade Tendermint to support this?

tomc974:
Yeah sure, I'll do that.

PR #196, opened 2018-04-05
add SimpleProofsFromMap

mossid:


merkle/simple_proof.go
```diff
@@ -22,6 +22,20 @@ func SimpleProofsFromHashers(items []Hasher) (rootHash []byte, proofs []*SimpleP
...
 }
 
+func SimpleProofsFromMap(m map[string]Hasher) (rootHash []byte, proofs []*SimpleProof) {
+	sm := NewSimpleMap()
+	for k, v := range m {
+		sm.Set(k, v)
+	}
+	sm.Sort()
+	kvs := sm.kvs
+	kvsH := make([]Hasher, 0, len(kvs))
```
melekes: why not preallocate array? `make([]Hasher, len(kvs))`

mossid: To make it not conflict with [`hashKVPairs`](https://github.com/tendermint/tmlibs/blob/master/merkle/simple_map.go#L78) in `SimpleMap.Hash`. It could work in that case too, however.

PR #199, opened 2018-04-13
Jae/fixprefixdb

jaekwon:
FEATURES:

 - [db] DebugDB shows better colorized output

BUG FIXES:

 - [db] PrefixDB Iterator/ReverseIterator fixes
 - [db] DebugDB fixes

### DIFF
--- a/db/mem_db.go
+++ b/db/mem_db.go
@@ -163,7 +163,7 @@ func (db *MemDB) ReverseIterator(start, end []byte) Iterator {
 	db.mtx.Lock()
 	defer db.mtx.Unlock()
 
-	keys := db.getSortedKeys(end, start, true)
+	keys := db.getSortedKeys(start, end, true)
 	return newMemDBIterator(db, keys, start, end)
 }
 
@@ -237,7 +237,8 @@ func (itr *memDBIterator) assertIsValid() {
 func (db *MemDB) getSortedKeys(start, end []byte, reverse bool) []string {
 	keys := []string{}
 	for key := range db.db {
-		if IsKeyInDomain([]byte(key), start, end, false) {
+		inDomain := IsKeyInDomain([]byte(key), start, end, reverse)
+		if inDomain {
 			keys = append(keys, key)
 		}
 	}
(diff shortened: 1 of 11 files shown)


db/mem_db.go
```diff
@@ -37,7 +37,8 @@ func (db *MemDB) Get(key []byte) []byte {
 	defer db.mtx.Unlock()
 	key = nonNilBytes(key)
 
-	return db.db[string(key)]
+	value := db.db[string(key)]
```
melekes: why not just return? copy?

db/prefix_db.go
```diff
@@ -201,48 +224,71 @@ func (pdb *prefixDB) Stats() map[string]string {
...
 	source Iterator
+	valid  bool
 }
 
-func newUnprefixIterator(prefix []byte, source Iterator) unprefixIterator {
-	return unprefixIterator{
-		prefix: prefix,
-		source: source,
+func newUnprefixIterator(prefix, start, end []byte, source Iterator) unprefixIterator {
+	if !source.Valid() || !bytes.HasPrefix(source.Key(), prefix) {
```
melekes: why not just

```
valid := !source.Valid() || !bytes.HasPrefix(source.Key(), prefix)
return unprefixIterator{
  ....
  valid: valid
}
```

db/prefix_db.go
```diff
@@ -201,48 +224,71 @@ func (pdb *prefixDB) Stats() map[string]string {
...
-	return itr.source.Valid()
+	return itr.valid && itr.source.Valid()
 }
 
 func (itr unprefixIterator) Next() {
+	if !itr.valid {
+		panic("unprefixIterator invalid, cannot call Next()")
+	}
 	itr.source.Next()
+	if !itr.source.Valid() || !bytes.HasPrefix(itr.source.Key(), itr.prefix) {
```
melekes: should probably be a function. used here and line 242

db/mem_db.go
jaekwon: No, just so it's easier to add print statements for debugging.  Or, I had to anyways, and this is an artifact.  I sometimes leave these around because it's convenient, and it doesn't seem like bad form.  How do you feel about it?

jaekwon: I've used delve before, that helps.  Print statements are still good too.  Have you tried Microsoft's Golang tooling?  I dunno if they have a debugger integrated.  What do you use?

PR #200, opened 2018-04-18
add MarshalJSON and UnmarshalJSON to BitArray

melekes:
Refs https://github.com/tendermint/tendermint/issues/795


jaekwon:
I don't think we should do this with MarshalJSON.
For developers, the existing struct's natural Amino:JSON format is fine.
Turning it into a custom formatted string makes it more difficult to create a compatible BitArray implementation in other languages.

common/bit_array.go
```diff
@@ -320,3 +322,58 @@ func (bA *BitArray) Update(o *BitArray) {
...
+// JSON description.
+func (bA *BitArray) UnmarshalJSON(b []byte) error {
+	if string(b) == "null" {
+		bA = nil
+		return nil
+	}
+
+	bAStr := string(b)
+
+	matches := bitArrayJSONRegexp.FindAllStringSubmatch(bAStr, 2)
```
jaekwon: There's no need to use All here, cuz the regexp is \A \z

common/bit_array.go
```diff
@@ -320,3 +322,58 @@ func (bA *BitArray) Update(o *BitArray) {
...
+	return []byte("\"" + bA.stringIndented("") + "\""), nil
+}
+
+var bitArrayJSONRegexp = regexp.MustCompile(`\A"BA{(\d+):([_X\s]+)}"\z`)
+
+// UnmarshalJSON implements json.Unmarshaler interface by unmarshaling a custom
+// JSON description.
+func (bA *BitArray) UnmarshalJSON(b []byte) error {
+	if string(b) == "null" {
+		bA = nil
```
jaekwon: This doesn't do anything.

ebuchman:
💃

PR #208, opened 2018-05-14
Move remotedb under db

ebuchman:
I also reran protoc with updated deps - that created a bunch of gross XXX code - hmm


ebuchman:
@cwgoes

cwgoes:
I get a different result for the compiled protobuf file:

https://gist.github.com/cwgoes/1ea6db7372a4a9c2b27dfb14d9b57f61

I wonder if we're using different `protoc` versions?

```bash
protoc --version

libprotoc 3.4.0
```

ebuchman:
```
ethanbuchman:cosmos-sdk$ protoc --version
libprotoc 3.4.0

```

ebuchman:
Could be different versions of protoc-gen-go ... it doesnt have a version flag/command tho

ebuchman:
I think I'll revert the updated proto for now ? 

What do you think of moving it all under db/remotedb ?

cwgoes:
Both proto files seem to work (at least, they pass the testsuite) - the previous generated file is like `types.pb.go` though, so might be worth staying uniform there.

I think `db/remotedb` makes sense, I just left the file structure from the original PR but I don't see any reason remotedb needs separate folders.

PR #210, opened 2018-05-22
tmhash

ebuchman:
@zmanian @Liamsi @jaekwon lets standardize around this ?

### DIFF
--- a/merkle/simple_map.go
+++ b/merkle/simple_map.go
@@ -2,7 +2,7 @@ package merkle
 
 import (
 	cmn "github.com/tendermint/tmlibs/common"
-	"golang.org/x/crypto/ripemd160"
+	"github.com/tendermint/tmlibs/merkle/tmhash"
 )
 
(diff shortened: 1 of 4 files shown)


Liamsi:
LGTM! Wouldn't it make more sense to have this code in go-crypto instead of tmlibs though?

ebuchman:
> Wouldn't it make more sense to have this code in go-crypto instead of tmlibs though?

Heh, fun question. We have an ongoing discussion about how to conceptualize/consolidate our repos.

Are you proposing that all of `merkle` move to `go-crypto` or just the new `tmhash` ? 

I wouldn't be opposed to moving all of `merkle`, but maybe then we'd want to move iavl as well ?!

ebuchman:
Actually there's already breaking stuff on develop here and I want to update tendermint to use this first. Rather than muck about with branches here let's just move it to go-crypto now: https://github.com/tendermint/go-crypto/pull/105

PR #211, opened 2018-05-22
dont length prefix hashes during concatenation

ebuchman:
Trying to simplify the tree here.

I don't see what the length prefixing buys us, other than another dependency/thing to specify.

Of course, we still need it for hashing kv pairs, otherwise two distinct pairs could hash to the same thing. 

Can we enforce that these FromXxxHashes functions only work with hashes of the same length (ie so theyre safe to concatenate without length prefixing ) ? 

@zmanian @jaekwon @Liamsi 

Ref https://github.com/tendermint/tendermint/issues/1547

### DIFF
--- a/merkle/simple_map.go
+++ b/merkle/simple_map.go
@@ -5,23 +5,27 @@ import (
 	"github.com/tendermint/tmlibs/merkle/tmhash"
 )
 
-type SimpleMap struct {
+// Merkle tree from a map.
+// Leaves are `hash(key) | hash(value)`.
+// Leaves are sorted before Merkle hashing.
+type simpleMap struct {
 	kvs    cmn.KVPairs
 	sorted bool
 }
 
-func NewSimpleMap() *SimpleMap {
-	return &SimpleMap{
+func newSimpleMap() *simpleMap {
+	return &simpleMap{
 		kvs:    nil,
 		sorted: false,
 	}
 }
 
-func (sm *SimpleMap) Set(key string, value Hasher) {
+// Hash the key and value and append to the kv pairs
+func (sm *simpleMap) Set(key string, value Hasher) {
 	sm.sorted = false
 
 	// Hash the key to blind it... why not?
-	khash := SimpleHashFromBytes([]byte(key))
+	khash := tmhash.Sum([]byte(key))
 
 	// And the value is hashed too, so you can
 	// check for equality with a cached value (say)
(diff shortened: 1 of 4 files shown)


melekes:
```
--- FAIL: TestSimpleMap (0.00s)
	Error Trace:	simple_map_test.go:31
	Error:      	Not equal: 
	            	expected: "acc3971eab8513171cc90ce8b74f368c38f9657d"
	            	actual  : "2db281ef2b0da1c20e16c9597b7cd529a9ce3c54"
	Test:       	TestSimpleMap
	Messages:   	Hash didn't match
	Error Trace:	simple_map_test.go:37
	Error:      	Not equal: 
	            	expected: "acc3971eab8513171cc90ce8b74f368c38f9657d"
	            	actual  : "2db281ef2b0da1c20e16c9597b7cd529a9ce3c54"
	Test:       	TestSimpleMap
	Messages:   	Hash didn't match
	Error Trace:	simple_map_test.go:44
	Error:      	Not equal: 
	            	expected: "0cd117ad14e6cd22edcd9aa0d84d7063b54b862f"
	            	actual  : "6c5244cc930cbbbe7ff83069e9d2a2ee3faf24ba"
	Test:       	TestSimpleMap
	Messages:   	Hash didn't match
	Error Trace:	simple_map_test.go:51
	Error:      	Not equal: 
	            	expected: "0cd117ad14e6cd22edcd9aa0d84d7063b54b862f"
	            	actual  : "6c5244cc930cbbbe7ff83069e9d2a2ee3faf24ba"
	Test:       	TestSimpleMap
	Messages:   	Hash didn't match
```

ebuchman:
Yeh the hashes are harcoded - need to update them. Just waiting for agreement first

Liamsi:
I do not think length prefixes are safer / neccessary either unless they encode the depth of where we are (which level) in the tree (which doesn't seem to be the case here). 

To prevent second pre-image attacks it is quite common to add a prefix depending on if we are hashing a leaf node, or and interior node. Is this intentional, that this doesn't seem to happen here?

ebuchman:
> Is this intentional, that this doesn't seem to happen here?

Not intentional, don't think we thought about it.

So the idea with a 2nd pre-image attack is to basically make a new merkle tree where the leaves are say the internal nodes from some layer in the original tree - so they have the same root but technically different leaves.

Semantically this shouldn't be a problem since the inner nodes will be meaningless from a tx/validator perspective, but its worth fixing this in the data structure so no one has to think about it. 

Let's copy CT then ? ie. prepend a 0x00 to the leaf hashes and a 0x1 to the inner hashes ?

ebuchman:
> Let's copy CT then ? ie. prepend a 0x00 to the leaf hashes and a 0x1 to the inner hashes ?

Oh, this won't go great in the implementation right now without some changes if we want to avoid using `append` (ie. performance reasons). The recursive impl doesnt distinguish leaves/nodes

ebuchman:
Looks like we can't really avoid append anyways. we can use `.Sum(prefix)` but that just uses append under the hood too. 

Need to be careful with this to make sure its done right and the same everywhere, especially since we're copying the merkle tree impl in a couple places in Tendermint as an optimization (to avoid allocations from the generic funcs in merkle pkg)

ebuchman:
Moved to https://github.com/tendermint/go-crypto/pull/105 , specifically commit https://github.com/tendermint/go-crypto/pull/105/commits/a72adddde5828174345a90fb0c8c8f879137fe2f

PR #213, opened 2018-05-24
use key without hashing, SimpleProofsFromMap returns map

mossid:


mossid:
I had to cache SimpleProofs of subtrees of RootMultiStore. With hashing the key, I had to traverse every possible combination to construct the map.

The comment says that it is for blinding the key, but leaf nodes are blinded anyway so paths don't expose keys. @ebuchman

ebuchman:
Moved to https://github.com/tendermint/go-crypto/pull/115

PR #214, opened 2018-05-24
no gogo proto

ebuchman:
Actually, do we need these to be protobuf defined in here? Would be great to just move them into the ABCI spec directly ... imports are annoying ...


melekes:
> Actually, do we need these to be protobuf defined in here? Would be great to just move them into the ABCI spec directly ... imports are annoying ...

agree. we should move

PR #217, opened 2018-06-04
flush on stop & function to close group

melekes:
as opposite to OpenGroup


ebuchman:
replace with https://github.com/tendermint/tmlibs/pull/218

PR #224, opened 2018-06-21
WIP - reverse iterators

jlandrews:
Seems to work, not sure if there's a better way to seek to last key for goleveldb.

Also needs more testing to check edge cases.

### DIFF
--- a/db/c_level_db.go
+++ b/db/c_level_db.go
@@ -190,7 +190,8 @@ func (db *CLevelDB) Iterator(start, end []byte) Iterator {
 }
 
 func (db *CLevelDB) ReverseIterator(start, end []byte) Iterator {
-	panic("not implemented yet") // XXX
+	itr := db.db.NewIterator(db.ro)
+	return newCLevelDBIterator(itr, start, end, true)
 }
 
 var _ Iterator = (*cLevelDBIterator)(nil)
@@ -204,7 +205,11 @@ type cLevelDBIterator struct {
 
 func newCLevelDBIterator(source *levigo.Iterator, start, end []byte, isReverse bool) *cLevelDBIterator {
 	if isReverse {
-		panic("not implemented yet") // XXX
+		if end != nil {
+			source.Seek(end)
+		} else {
+			source.SeekToLast()
+		}
 	}
 	if start != nil {
 		source.Seek(start)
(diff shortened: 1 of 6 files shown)


xla:
@jlandrews This repo is going to collapse into tendermint proper. Which should be done by tomorrow, can you resubmit your PR there?

db/c_level_db.go
```diff
@@ -204,7 +205,11 @@ type cLevelDBIterator struct {
 
 func newCLevelDBIterator(source *levigo.Iterator, start, end []byte, isReverse bool) *cLevelDBIterator {
 	if isReverse {
-		panic("not implemented yet") // XXX
+		if end != nil {
```
jaekwon: tmlibs/db/types.go docs: 
```go
    // Iterate over a domain of keys in descending order. End is exclusive.
    // Start must be greater than end, or the Iterator is invalid.
    // If start is nil, iterates from the last/greatest item (inclusive).
    // If end is nil, iterates up to the first/least item (inclusive).
    // CONTRACT: No writes may happen within a domain while an iterator exists over it.
    // CONTRACT: start, end readonly []byte
    ReverseIterator(start, end []byte) Iterator
```
So we want:

```go
if start == nil {
    if isReverse {
        source.SeekToLast()
    } else {
        source.SeektoFirst()
    }
} else {
    source.Seek(start)
}
```

db/c_level_db.go
```diff
@@ -243,7 +248,12 @@ func (itr cLevelDBIterator) Valid() bool {
 	// If key is end or past it, invalid.
 	var end = itr.end
 	var key = itr.source.Key()
-	if end != nil && bytes.Compare(end, key) <= 0 {
+	if !itr.isReverse && end != nil && bytes.Compare(end, key) <= 0 {
+		itr.isInvalid = true
+		return false
+	}
+
+	if itr.isReverse && start != nil && bytes.Compare(start, key) > 0 {
```
jaekwon: I would opt for if {} else {} for clarity regarding the isReverse check.

jlandrews:
Comments addressed. Still plan to add more extensive testing to catch edge cases.

db/go_level_db.go
```diff
@@ -207,10 +208,13 @@ type goLevelDBIterator struct {
 var _ Iterator = (*goLevelDBIterator)(nil)
 
 func newGoLevelDBIterator(source iterator.Iterator, start, end []byte, isReverse bool) *goLevelDBIterator {
-	if isReverse {
-		panic("not implemented yet") // XXX
+	if isReverse && start == nil {
+		for !source.Last() {
```
jaekwon: This doesn't seem right.... I'll submit a fix to this branch.  https://github.com/syndtr/goleveldb/blob/master/leveldb/iterator/iter.go#L29

jlandrews: Don't know why I thought Last was just checking and not seeking... Thanks for catching that.

